PR #220, opened 2014-04-07
Fix gemspec and run transpec

petergoldstein:
1. Fixed gemspec to load by using 'activesupport'
2. Updated gemspec to allow more recent versions of RSpec
3. Ran transpec to update specs to RSpec 3.x syntax

### DIFF
--- a/ice_cube.gemspec
+++ b/ice_cube.gemspec
@@ -18,7 +18,7 @@ Gem::Specification.new do |s|
   s.rubyforge_project = "ice-cube"
 
   s.add_development_dependency('rake')
-  s.add_development_dependency('rspec', '~> 2.12.0')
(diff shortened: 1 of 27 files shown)


seejohnrun:
Obviously it's been a while here, but `expect` is definitely the norm now (at least in my world).  If you are willing to re-run this I'd love to get it merged, and sorry for doubting it!

petergoldstein:
I don't think a 2+ year old PR is a good basis for a patch, so I'm closing this PR.

I don't use this gem anymore, so I'm not the best choice for coming up with a replacement PR.  But you should be able to run transpec on an updated master and get a correctly updated result.

seejohnrun:
Cool thanks Peter

PR #259, opened 2014-10-06
Use I18n and add Japanese locale.

tomoyuki28jp:
I made changes based on the work made by @joelmeyerhamme ( https://github.com/joelmeyerhamme/ice_cube/tree/international ).
https://github.com/seejohnrun/ice_cube/issues/68

PR #273, opened 2014-12-09
Internationalization: fixing tests in pull #259 from @tomoyuki28jp

dgilperez:
Hi! 

I'm creating a new PR, since I don't know how to add a commit to #259. I suspect the failing tests I mentioned earlier are failing only in my machine and I'd like to see what does Travis think about that. Please feel free to delete this PR if this is not the proper way of doing it. 

Regards!

### DIFF
--- a/lib/ice_cube.rb
+++ b/lib/ice_cube.rb
@@ -71,7 +71,7 @@ module IceCube
   # Defines the format used by IceCube when printing out Schedule#to_s.
   # Defaults to '%B %e, %Y'
   def self.to_s_time_format
-    @to_s_time_format ||= I18n.t("ice_cube.date.formats.default")
+    I18n.t("ice_cube.date.formats.default")
   end
 
   # Sets the format used by IceCube when printing out Schedule#to_s.
--- a/lib/ice_cube.rb
+++ b/lib/ice_cube.rb
@@ -71,7 +71,7 @@ module IceCube
   # Defines the format used by IceCube when printing out Schedule#to_s.
   # Defaults to '%B %e, %Y'
   def self.to_s_time_format
-    @to_s_time_format ||= '%B %e, %Y'
+    @to_s_time_format ||= I18n.t("ice_cube.date.formats.default")
   end
 
   # Sets the format used by IceCube when printing out Schedule#to_s.
--- a/lib/ice_cube.rb
+++ b/lib/ice_cube.rb
@@ -1,5 +1,8 @@
 require 'date'
 require 'ice_cube/deprecated'
+require 'i18n'
+
+I18n.load_path += Dir[File.expand_path('../../config/locales/*{rb,yml}', __FILE__)]
 
 module IceCube
 
--- a/lib/ice_cube/builders/ical_builder.rb
+++ b/lib/ice_cube/builders/ical_builder.rb
@@ -32,15 +32,15 @@ module IceCube
 
     def self.ical_utc_format(time)
       time = time.dup.utc
-      "#{time.strftime('%Y%m%dT%H%M%SZ')}" # utc time
+      I18n.l(time, format: '%Y%m%dT%H%M%SZ') # utc time
     end
 
     def self.ical_format(time, force_utc)
       time = time.dup.utc if force_utc
       if time.utc?
-        ":#{time.strftime('%Y%m%dT%H%M%SZ')}" # utc time
+        ":#{I18n.l(time, format: '%Y%m%dT%H%M%SZ')}" # utc time
       else
-        ";TZID=#{time.strftime('%Z:%Y%m%dT%H%M%S')}" # local time specified
+        ";TZID=#{I18n.l(time, format: '%Z:%Y%m%dT%H%M%S')}" # local time specified
       end
     end
 
--- a/lib/ice_cube/builders/string_builder.rb
+++ b/lib/ice_cube/builders/string_builder.rb
@@ -34,9 +34,6 @@ module IceCube
 
     module Helpers
 
-      NUMBER_SUFFIX = ['th', 'st', 'nd', 'rd', 'th', 'th', 'th', 'th', 'th', 'th']
-      SPECIAL_SUFFIX = { 11 => 'th', 12 => 'th', 13 => 'th', 14 => 'th' }
-
       # influenced by ActiveSupport's to_sentence
       def sentence(array)
         case array.length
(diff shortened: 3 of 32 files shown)


dgilperez:
Any chance this PR gets merged anytime soon? Thanks!

stravabr:
Yes please! :+1:

dgilperez:
Just checking in to keep this thing alive. @avit, it would be nice if you could have a look at it, or bring in some of the other maintainers. Thanks!

isaacseymour:
Awesome work @dgilperez ! Would love to see this merged

lib/ice_cube.rb
```diff
@@ -68,7 +71,7 @@ module Validations
   # Defines the format used by IceCube when printing out Schedule#to_s.
   # Defaults to '%B %e, %Y'
   def self.to_s_time_format
-    @to_s_time_format ||= '%B %e, %Y'
+    I18n.t("ice_cube.date.formats.default")
```
isaacseymour: This eliminates the option of overriding the format from the i18n config
```ruby
@to_s_time_format ||= I18n.t("ice_cube.date.formats.default")
```
maybe?

lib/ice_cube/builders/ical_builder.rb
```diff
@@ -32,15 +32,15 @@ def to_s
 
     def self.ical_utc_format(time)
       time = time.dup.utc
-      "#{time.strftime('%Y%m%dT%H%M%SZ')}" # utc time
+      I18n.l(time, format: '%Y%m%dT%H%M%SZ') # utc time
```
isaacseymour: Probably shouldn't use `I18n` here because it's a non-localised time format?

lib/ice_cube/builders/string_builder.rb
```diff
@@ -13,14 +13,18 @@ def piece(type, prefix = nil, suffix = nil)
     end
 
     def to_s
-      @types.each_with_object(@base || '') do |(type, segments), str|
+      string = @base || ''
+      @types.each do |type, segments|
```
isaacseymour: This could be done nicely with `reduce`:
```ruby
@types.reduce(@base || '') do |string, (type, segments)|
  if formatter = self.class.formatter(segments)
    current = formatter.call(segments)
  else
    next if segments.empty?
    current = self.class.sentence(segments)
  end

  format = I18n.t('ice_cube.string.format')[type] ? type : 'default'
  I18n.t("ice_cube.string.format.#{format}"), rest: string, current: current)
end
```

lib/ice_cube.rb
dgilperez: You could always change `I18n.t("ice_cube.date.formats.default")` to override that :smile:, that is actually my proposed way to do that. What do you think? But maybe there are some docs / other methods to be changed accordingly?

lib/ice_cube/builders/ical_builder.rb
dgilperez: Agreed. Let me change that.

lib/ice_cube/builders/string_builder.rb
dgilperez: On the way ...

avit:
I will make the time to review this soon. This does look like a helpful addition.

Currently IceCube has no runtime dependencies outside of stdlib. (This change would introduce the I18n gem, which I'm OK with since it doesn't have a whole tree of other dependencies.) Before I look at your implementation, will it be possible to make translations optional? (e.g. require "ice_cube/i18n")

(Sorry this hasn't been a priority, I don't use the string output for schedule descriptions myself; I only use it for generating occurrence times.)

avit:
And thanks for your work and patience, everyone.

lib/ice_cube.rb
isaacseymour: Yeah, actually caching is a really bad idea if you're using multiple locales in the same process (e.g. tests) - the second locale's date format gets ignored. My bad.

lib/ice_cube/builders/string_builder.rb
dgilperez: Hey, I can't figure out why your changes are not producing the same test results ... and have no time to spare on this for a few days. Would you add to the PR yourself?

dgilperez:
No pro @avit! Thanks for having a look at this. I will have no spare time in the following few days, so I suggest you adapt the code at this PR to the current codebase, add that option to enable / disable translations and I'll get back as soon as possible!

PS: BTW, I don't think enabling / disabling translations is a good idea, or at least I don't see the use case. I'd rather add the defaults to the english `.yml` version and build from that. Current users should be able to upgrade without noticing.

lib/ice_cube/builders/string_builder.rb
isaacseymour: Would love to but I don't have push rights on your fork :P I'll add them to gocardless/ice_cube (along with French translation) and PR them into your fork.

dgilperez: Sounds good to me!!!

On Tue, Mar 17, 2015 at 11:00 AM, Isaac Seymour <notifications@github.com>
wrote:

> In lib/ice_cube/builders/string_builder.rb
> <https://github.com/seejohnrun/ice_cube/pull/273#discussion_r26560481>:
>
> > @@ -13,14 +13,18 @@ def piece(type, prefix = nil, suffix = nil)
> >      end
> >
> >      def to_s
> > -      @types.each_with_object(@base || '') do |(type, segments), str|
> > +      string = @base || ''
> > +      @types.each do |type, segments|
>
> Would love to but I don't have push rights on your fork :P I'll add them
> to gocardless/ice_cube (along with French translation) and PR them into
> your fork.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/seejohnrun/ice_cube/pull/273/files#r26560481>.
>



-- 
David Gil Pérez
qoolife.com
tel: +34 637 896 543 | skype: dgilperez | @dgilperez
<http://twitter.com/dgilperez>

mlerner:
@avit @dgilperez Looks great! Would be nice to get this merged soon :)

stravabr:
:+1:

jpowell:
:+1: was looking for a way to get this from `I18n` also.

dgilperez:
So, @avit, any news?

dgilperez:
Hey @seejohnrun, are there any plans to merge this PR? Anything left to do here?

dgilperez:
Closing due to #311 been merged.

PR #284, opened 2015-03-10
Fix WEEKLY recurring when specifying non default WKST

willemave:
fixes a nasty edge case where we advance past the start of the opening_time and miss occurrences.


mitchel456:
Any chance of getting this reviewed/merged? We're currently using a pretty outdated fork that fixes this issue, but we'd love to be able to use the most current version.

pocman:
+1

MGPalmer:
+1 Except I found a problem with this - it seems to cause biweekly occurrences which start on a sunday, with weekstart set to monday, to miss their first occurrence X)

avit:
This still needs the logic to realign the wday start. Please see updated #383

PR #290, opened 2015-05-28
Adding namespacing to Rule in Readme to make it easier to run examples

compwron:
While experimenting with this library with my team I ran into the inconvenience that some of the usages of Rule in the readme are namespaced to IceCube:: and some are not, so here is a PR. :space_invader:

### DIFF
--- a/README.md
+++ b/README.md
@@ -43,7 +43,7 @@ With ice_cube, you can specify (in increasing order of precedence):
 * Recurrence Times - To specifically include in a schedule
 * Exception Times - To specifically exclude from a schedule
 
-Example: Specifying a recurrence with an exception time (ActiveRecord:
+Example: Specifying a recurrence with an exception time
 
 ```ruby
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -42,11 +43,11 @@ With ice_cube, you can specify (in increasing order of precedence):
 * Recurrence Times - To specifically include in a schedule
 * Exception Times - To specifically exclude from a schedule
 
-Example: Specifying a recurrence with an exception time:
+Example: Specifying a recurrence with an exception time (ActiveRecord:
```
seejohnrun: what's this one doing?

compwron: Oops. That was going to grow into a 'require' because 1.day etc only works with rails helpers but then I went back to take it out and missed one.

seejohnrun: Cool - and thanks so much for the PR!
If we can get that fixed I'm excited to merge!

compwron:
Awesome! Thanks :)

seejohnrun:
Thank you!

PR #295, opened 2015-06-29
Parse time zones when creating schedules from ICAL; fix DST test bug

barelyknown:


barelyknown:
I'm not seeing the CI errors in my test environment. I'll look closer.

johnhamelink:
I'd love to see this merged - what's stopping it from getting included?

johnhamelink:
FYI I found a bug in that PR which I fixed here: https://github.com/togglepro/ice_cube/pull/1 - I will add a test for it to prove it's fixed shortly.

WojtekKruszewski:
Without time zone it falls back to Time.parse which uses system time zone, ignoring zone configured in ActiveSupport. IceCube docs mention using ActiveSupport time zone support, so would be great to try Time.zone.parse first, and then fall back to Time.parse.

dcosson:
What's blocking this from being merged? This is a really bad bug, the `from_ical` method imports things at the wrong time whenever a timezone is used in the ical (which is completely legit by the standard)

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -23,6 +25,14 @@ def self.schedule_from_ical(ical_string, options = {})
       Schedule.from_hash data
     end
 
+    def self._parse_in_tzid(value, tzid)
+      time_parser = Time
+      if tzid
+        time_parser = ActiveSupport::TimeZone.new(tzid.split('=')[1]) || Time
```
dcosson: This is actually inconsistent with ice_cube's behavior for `to_ical`. Ice_cube's to_ical uses the `%Z` strftime string which (at least for en_US based locale's, behavior may be different elsewhere) outputs a string like `PDT` which is effectively a utc offset not a timezone (PDT= UTC - 7 hours vs PST = UTC - 8 hours).

But `ActiveSupport::TimeZone.new` requires an actual timezone, either the Olson format timezone such as `America/Los_Angeles` or the other version of `Pacific Time (US & Canada)` (which is some US only format that I don't even know if it's a standard with a name). Furthermore, there's not an ActiveSupport method for going from a timezone offset like PDT to a region's timezone because they're not the same thing.

So, in order for this PR to really make sense in the library, the `to_ical` method should be changed to use the `.time_zone.name` property of ActiveSupport::TimeWithZone instead of the strftime string.

I checked the ical [RFC](https://tools.ietf.org/html/rfc5545#section-3.2.19) and it does not specify timezone identifiers, but says `Implementers may want to use the naming conventions defined in existing time zone specifications such as the public-domain TZ database [TZDB]` and links to the olson [timzone database](http://www.twinsun.com/tz/tz-link.htm)

Tybot204:
Any new news regarding getting this PR updated and merged? I just ran into this problem in a project of mine creating an event system. I was about to fork the repo but found this PR does what I need!

avit:
Closing this to keep discussion on the more recent #335. However, ical parsing is something that we might consider moving out of this gem, see #382 for discussion.

Issue #303, opened 2015-10-06
iCal parsing doesn't process multiple rules


sethdeckard:
iCal parsing is only handling the last rule processed, example:

``` ruby
schedule = IceCube::Schedule.new
schedule.add_recurrence_rule IceCube::Rule.weekly.day(:monday, :tuesday)
schedule.add_recurrence_rule IceCube::Rule.weekly(2).day(:friday)

cal = schedule.to_ical
=> "DTSTART;TZID=CDT:20151005T195541\nRRULE:FREQ=WEEKLY;BYDAY=MO,TU\nRRULE:FREQ=WEEKLY;INTERVAL=2;WKST=SU;BYDAY=FR"

parsed = IceCube::Schedule.from_ical(cal)
parsed.recurrence_rules.size
=> 1

cal2 = parsed.to_ical
=> "DTSTART;TZID=CDT:20151005T195541\nRRULE:FREQ=WEEKLY;INTERVAL=2;WKST=SU;BYDAY=FR"
```

Morantron:
:+1: just jumped into this issue

dimerman:
:+1:

dimerman:
I have this https://github.com/seejohnrun/ice_cube/pull/308 for your consideration.

PR #304, opened 2015-10-06
Fxes issue #303 preventing multiple rule parsing

sethdeckard:
See issue #303 for example, just a simple fix to add elements to an array in the hash key rather than assigning a new array every time a rule was processed.


joelfperion:
This looks good, why is it not merged?

dimerman:
:+1:  why indeed?

avit:
We merged #308 instead.

PR #307, opened 2015-10-18
Add validation to prevent nil rules

sethdeckard:
Stumbled on this by accident but the `add_recurrence_rule` / `add_exception_rule` methods allow nil which later leads to errors like below when invoking various methods like `occurrences_between`:

`NoMethodError: undefined method`reset' for nil:NilClass`

Simple example:

``` ruby
s = IceCube::Schedule.new(Time.now)
s.add_recurrence_rule(Random.new)
s.occurrences_between(Time.now, Time.now + 36000)
```

I was also thinking it could be a good idea to add a `is_a?(Rule)` validation as well since there are quite a few methods an object has to respond to in order to work as a Rule but I just left it at a simple nil check for now.

### DIFF
--- a/lib/ice_cube/schedule.rb
+++ b/lib/ice_cube/schedule.rb
@@ -68,6 +68,7 @@ module IceCube
 
     # Add a recurrence rule to the schedule
     def add_recurrence_rule(rule)
+      fail ArgumentError, 'Rule cannot be nil' if rule.nil?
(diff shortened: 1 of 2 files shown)


avit:
I like this, but it looks like we've changed it to ignore nil values in the meantime... we could still change it to raise the error instead (which seems to make more sense).

I think the same should apply to adding single times as well.

PR #308, opened 2015-10-27
fixed parsing of ical with multiple recurrence rules

dimerman:
Each new rule parsed from ical overwrote earlier ones.


avit:
Thanks for the fix!

PR #313, opened 2015-12-17
Parse incoming RDATE lines from iCal files, handle all-day RDATE and EXDATE

lilole:
Howdy friends. I have been evaluating different options to enumerate occurrences between dates, given only an iCal file. I think ice_cube is great for it. However, I must support RDATE, and it seems this change is all I need to finally adopt ice_cube. I hope you find it useful, and it finds its way into the gem. Cheers.

### DIFF
--- a/lib/ice_cube/builders/ical_builder.rb
+++ b/lib/ice_cube/builders/ical_builder.rb
@@ -35,12 +35,20 @@ module IceCube
       IceCube::I18n.l(time, format: '%Y%m%dT%H%M%SZ') # utc time
     end
 
-    def self.ical_format(time, force_utc)
+    def self.ical_format(time, force_utc, rule=nil)
       time = time.dup.utc if force_utc
-      if time.utc?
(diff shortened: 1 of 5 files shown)


dan-higgins:
I know the branch says "3_lines", but I needed to support all-day events for RDATE and EXDATE, so this version does that.

It also allows for some "extensions" on parsing iCal lines, including (a) blank lines, and (b) comment lines starting with "#", and (c) if a line looks like a date "yyyymmdd", treat it as an RDATE automatically.

These extensions are all disabled by default, but can be enabled with `IceCube::IcalParser.use_extensions = true`, or setting env var `ICE_CUBE_ICAL_EXTS` to non-empty.

I know this is a lot, but if you are interested in merging, I will be happy to add rspecs for all of this. I just needed it for work and figured it would be good to share.  Cheers.

lilole:
I think the changes here have been superseded by others in the past 5 years.  :slightly_smiling_face:

PR #318, opened 2016-01-28
deprecation raised if start_time is missing

pocman:
IceCube: :start_date is deprecated, please use :start_time in `to_schedule' should not be raised if :start_time is present.


avit:
I fixed this by making `IceCube.compatibility` match he released version number. "start_date" was still there to output for the older serialized format, but that's unsupported now.

Thanks for the contribution though!

PR #320, opened 2016-01-31
I18n cleanup

dnrce:
Awesome work getting i18n merged!

Here's a little cleanup of the i18n YAML:
- Use `date.day_names` and `date.month_names` keys directly rather than duplicating them under the `ice_cube` key
- Remove unused keys
  - `date.formats`
  - `date.order`
  - `time`
- Make whitespace consistent across files, and consistent with Ruby's `to_yaml`


dnrce:
My reference was rails-i18n and I noticed that the Japanese `day_names` in ice_cube don't match. Rails-i18n uses e.g. `水曜日` whereas ice_cube has only `水曜`. I know no Japanese so I left it as it already was in ice_cube.

avit:
Not sure why we didn't get to review and merge this sooner, but thank you! I rebased and merged this now.

PR #321, opened 2016-01-31
Travis: Update bundler before installing gems

dnrce:
Some versions of bundler are affected by a bug that causes them to throw

    NoMethodError: undefined method `spec' for nil:NilClass

during installation of the yaml_db gem. This bug was fixed in bundler@a0c90cd, but several of
Travis's Ruby environments are based on older affected versions. To resolve, we force an update.


dnrce:
Fixed in #328

PR #325, opened 2016-02-19
Remove deprecation warning

pocman:
We don't want to be warn if the `start_date` key is only present for retrocompatibility purpose.


avit:
Support for `start_date` will be removed. If you're still getting this warning with the current version then you need to make sure you update your hash input before it breaks.

PR #326, opened 2016-02-21
Fix day_of_year

### LINKED ISSUE #277: ice_cube not selecting the last day of the year correctly?
I was testing with this gem and noticed when I used:
# every fourth year on new year's eve

schedule.add_recurrence_rule Rule.yearly(4).day_of_year(-1)

the date returned was 12/30 and not 12/31.

my example:
[1] pry(main)> require 'ice_cube'
=> true
[2] pry(main)> last_day_of_year = IceCube::Schedule.new(now = Time.now) do |s|
[2] pry(main)\*   s.add_recurrence_rule(IceCube::Rule.yearly.day_of_year(-1))  
[2] pry(main)\* end
=> #<IceCube::Schedule:0x00000002784590
 @all_exception_rules=[],
 @all_recurrence_rules=
  [#<IceCube::YearlyRule:0x000000023746f8
    @interval=1,
    @uses=0,
....
[3] pry(main)> last_day_of_year.next_occurrence
=> 2015-12-30 13:54:49 -0600

In order to get the 31st, I used:
   s.add_recurrence_rule(IceCube::Rule.yearly.day_of_year(0))

Am I doing something wrong?
Thank you!

mitchel456:
Looks like a simple fix - i.e. `day_of_year(-1)` should refer to day 365 (in a non-leap year), but as written it'll result in 364 (-1 + 365, on line 34). This should fix #277.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,5 @@
+before_install:
+  - gem update bundler
 script: "bundle exec rspec spec"
 notifications:
   email:
--- a/spec/examples/hourly_rule_spec.rb
+++ b/spec/examples/hourly_rule_spec.rb
@@ -42,10 +42,10 @@ module IceCube
         schedule = Schedule.new(t0 = Time.local(2013, 11, 3, 0, 0, 0))
         schedule.add_recurrence_rule Rule.hourly
         schedule.first(4).should == [
-          Time.local(2013, 11, 3, 0, 0, 0),             # -0700
-          Time.local(2013, 11, 3, 1, 0, 0) - ONE_HOUR,  # -0700
-          Time.local(2013, 11, 3, 1, 0, 0),             # -0800
-          Time.local(2013, 11, 3, 2, 0, 0),             # -0800
+          Time.new(2013, 11, 3, 0, 0, 0, '-07:00'),
+          Time.new(2013, 11, 3, 1, 0, 0, '-07:00'),
+          Time.new(2013, 11, 3, 1, 0, 0, '-08:00'),
+          Time.new(2013, 11, 3, 2, 0, 0, '-08:00')
         ]
       end
 
--- a/lib/ice_cube/validations/day_of_year.rb
+++ b/lib/ice_cube/validations/day_of_year.rb
@@ -31,7 +31,7 @@ module IceCube
 
       def validate(step_time, schedule)
         days_in_year = TimeUtil.days_in_year(step_time)
-        yday = day < 0 ? day + days_in_year : day
+        yday = day < 0 ? day + days_in_year + 1 : day
         offset = yday - step_time.yday
         offset >= 0 ? offset : offset + days_in_year
       end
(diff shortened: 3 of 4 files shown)


.travis.yml
```diff
@@ -1,3 +1,5 @@
+before_install:
+  - gem update bundler
```
mitchel456: Stole this from https://github.com/seejohnrun/ice_cube/pull/321 - that build continued to fail because of an unrelated issue, addressed in this PR.

spec/examples/hourly_rule_spec.rb
```diff
@@ -42,10 +42,10 @@ module IceCube
         schedule = Schedule.new(t0 = Time.local(2013, 11, 3, 0, 0, 0))
         schedule.add_recurrence_rule Rule.hourly
         schedule.first(4).should == [
-          Time.local(2013, 11, 3, 0, 0, 0),             # -0700
-          Time.local(2013, 11, 3, 1, 0, 0) - ONE_HOUR,  # -0700
-          Time.local(2013, 11, 3, 1, 0, 0),             # -0800
-          Time.local(2013, 11, 3, 2, 0, 0),             # -0800
+          Time.new(2013, 11, 3, 0, 0, 0, '-07:00'),
```
mitchel456: I'm pretty sure this test as written relied on incorrect behavior of `Time::local` - unless I'm really confused (always possible) 1:00 AM on November 3rd, 2013, is still in DST, as DST doesn't end until 2:00 AM. I've filed this issue against Ruby as https://bugs.ruby-lang.org/issues/12147. At any rate, I think this rewritten test is a bit of an improvement, as it makes the offset an explicit part of the test.

mitchel456: A coworker pointed out to me that I am, indeed, confused - there are two instances of 1:00 AM on Nov 3rd, 2013, and `Time::local` has to pick one. I'd expect it to be the first one, but picking the second one isn't necessarily incorrect behavior. Because of this ambiguity, though, I still think the best solution is to write these tests explicitly, so that Ruby's semi-arbitrary choice of which 1:00 AM to return doesn't affect them.

avit: Time.local and Time.new are the same thing: `local` is just a little more explicit so we prefer it when we don't mean `utc`.

Yes, it's the second hour. I think it makes sense since most of Nov 3 is Standard Time, and if you want to specify an hour of that day it should be the expected offset. Only 00:00 to 00:59 are DST.

The `- ONE_HOUR` is a trick to "rewind" from 1AM Standard Time into 1AM DST because there's no other way to specify that "extra" hour literally. (I guess there's `Time.at unixtime` too, though.)

avit:
The first commit on this branch looks good at a glance: I can apply the bundler changes separately, and the last commit with the DST issue isn't valid.

I guess this problem has gone unnoticed until this year because 2016 is a leap year, and a specific test for "Dec 31" as "-1" was untested for regular years before... can we add a test to verify both leap years and non-leap years? The scenario "every fourth year on new year's eve" from #277 seems like a worthwhile scenario to specify here.

avit:
Merged in #381. Thanks!

PR #329, opened 2016-02-23
Rules can override schedule's duration

### LINKED ISSUE #322: Moving duration (or end_time) to Rules instead of Schedule
Thanks for this gem, it's super useful and I like everything.

There's a feature I would like to implement though, but before starting I'd ask you what you think about it.

I suppose this gem is built with the convintion that the duration (or end_time) should be related to the schedule, but I think it's better to assign duration to the Rules. I explain why.

I'm creating a Rails booking engine, that should allow developers to implement a booking system for various use cases (hotels, apartments, shows, meeting rooms etc...). Let's take the MeetingRoom case.
I would create an IceCube::Schedule attribute into the MeetingRoom model, that represents the availability of the MeetingRoom. Let's say this MeetingRoom is available on mondays and tuesdays from 8AM to 10AM and from 2PM to 6PM.

the 8AM-10AM interval has a duration of 2 hours, while the 2PM to 6PM has a duration of 4 hours. Having the duration related to Rules would allow for keeping the whole availability of the MeetingRoom inside the same Schedule. This would be great for making occurrence queries once, instead of having a number schedules I should cycle everytime. 

I hope I explained the problem. I'm really interested in this gem, and it would be great to include it in my engine, but the single schedule thing is a costraint I need.

Chosko:
fixes #322.

I added the possibility to override the duration of a schedule in a rule. This is made with a new validator (OverrideDuration) to keep the syntax as easy as `Rule.daily.override_duration(4.hours)`

I changed all the methods that make use of `schedule.duration` to give precedence to `rule.duration`

**About builders**

* `StringBuilder` was not affected because I see that `schedule.to_s` doesn't output schedule's duration either
* `HashBuilder` is used to store `:duration` in case the rule has a duration.
* `ICalBuilder` was not changed because adding 'DURATION' would bring to conflicts with schedule's duration or even with other rules. Maybe it's not an optimal solution, but I didn't see in the code a handling of the iCal property 'DURATION', so I didn't want to break anything.

I tested with these ruby versions: 1.9.3-p551, 2.0.0-p643, ruby-2.1.5, ruby-2.2.2, ruby-2.3.0 1.9.3

I didn't change or remove any old test and I just added new tests to fully cover all my changes. All tests pass in my machine.

Hope this may help

### DIFF
--- a/lib/ice_cube.rb
+++ b/lib/ice_cube.rb
@@ -43,6 +43,7 @@ module IceCube
 
     autoload :Count, 'ice_cube/validations/count'
     autoload :Until, 'ice_cube/validations/until'
+    autoload :OverrideDuration, 'ice_cube/validations/override_duration'
 
     autoload :SecondlyInterval, 'ice_cube/validations/secondly_interval'
     autoload :MinutelyInterval, 'ice_cube/validations/minutely_interval'
--- a/lib/ice_cube/rule.rb
+++ b/lib/ice_cube/rule.rb
@@ -11,6 +11,11 @@ module IceCube
       until_time || occurrence_count
     end
 
+    # Does this rule override the schedule's duration?
+    def overrides_duration?
+      !duration.nil?
+    end
+
     def ==(rule)
       if rule.is_a? Rule
         hash = to_hash
--- a/lib/ice_cube/schedule.rb
+++ b/lib/ice_cube/schedule.rb
@@ -44,6 +44,16 @@ module IceCube
       @end_time = start_time + seconds
     end
 
+    # Return the maximum value between schedule's duration and
+    # the durations of all the recurrence and exception rules included
+    # in this schedule
+    def max_duration
+      all_durations = [duration]
+      all_durations += @all_recurrence_rules.map { |r| r.duration }.compact
+      all_durations += @all_exception_rules.map { |r| r.duration }.compact
+      all_durations.max
+    end
+
     # Add a recurrence time to the schedule
     def add_recurrence_time(time)
       return nil if time.nil?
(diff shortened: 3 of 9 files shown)


Chosko:
@seejohnrun I don't know what's going on with Travis, But all tests pass on my machine with ruby >= 1.9.3 (up to 2.3.0)

seejohnrun:
@Chosko can you give rebasing a shot?

seejohnrun:
@Chosko fixed some travis issues this morning

Chosko:
@seejohnrun I rebased on latest master commit. :)

seejohnrun:
Cool - thank you (and sorry for the delay I was on vacation)

My one remaining issue with merging this is that it adds something that iCal doesn't support - which we haven't done so far because then it won't be compatible with an iCal export.

What are your thoughts on that?

avit:
I commented on this question in #322 @seejohnrun

Chosko:
@seejohnrun, @avit  Sorry for my late answer.
I think I misunderstood the way ice_cube maps Schedules and Rules to iCal, and I did notice it when I already finished the implementation. Now that I read better the RFC I understand my error.

A solution like https://github.com/seejohnrun/ice_cube/issues/131 would provide a good solution.
I'm sorry to say that I have no more time to go further into this feature. 

I'll use my fork now because it suits my need, but I'll definitely move back to the original ice_cube as soon as someone implements a better solution. Feel free to close it withouth merging.

seejohnrun:
Thanks @Chosko for the great work (I truly appreciate the time you put in) - we'll get something out soon!

PR #331, opened 2016-02-25
Fix february in readme example

fwininger:


fwininger:
@seejohnrun could you do a review on it ?

Thanks!

PR #343, opened 2016-06-20
iCal spec defines UNTIL as either a date or date-time.

figwit:
Support for rules with UNTIL part set as a date, avoiding coercion into a Time in the system time zone.

https://tools.ietf.org/html/rfc5545#section-3.3.10


stevesims:
+1

davegudge:
+1

ZachGagnon1:
+1

PR #347, opened 2016-07-27
Fix for bug occurring with biweekly schedules with week start set to monday

MGPalmer:
Hello!

We've run into this problem in our production code (was quite a journey to track down that on certain days, the next scheduled occurrence can be "skipped" due to this bug). We're not 100% the fix is not breaking anything or is exhaustive, however, it passes all our and this gem's specs.

Oh BTW, many many thanks and respect for providing this gem! It's quite an integral part of our business...


MGPalmer:
Nope, this "fix" causes a different issue (Schedules with monday weekstart, occurring biweekly on sundays get displayed wrongly every week).

PR #349, opened 2016-08-23
Support BYSETPOS for MONTHLY AND YEARLY freq

NicolasMarlier:
Self-explanatory ;)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ coverage
 coverage.data
 *.gem
 .bundle
+.idea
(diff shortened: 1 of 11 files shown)


avit:
Interesting! This option doesn't seem to be clearly documented in the icalendar RFC but it looks like it's widely supported by different tools. I believe it would be worth finishing this up and adding it.

http://www.kanzaki.com/docs/ical/recur.html

pacso:
I believe this was implemented in #449

PR #353, opened 2016-10-04
Allow serialization of single occurrence rule

mintuhouse:


avit:
Hi!

Sorry for the late reply on this, but I'm not sure why we need to serialize this. The "single occurrence rule" is internal only, and is implied from the schedule start time when there are no recurrence rules. (A serialized schedule with only a start time will always be reloaded with a single occurrence rule.)

I'm closing this, but if you can describe a good reason for needing this, we may reopen it.

PR #358, opened 2016-10-21
German locale, testing

joelmeyerhamme:
Thought I should add a German locale after I started the I18n a few years ago 😀. It was great, very easy to do.


seejohnrun:
Thank you, thanks for the original and thanks for the tests (which are good to serve as a base check of i18n in a non-english language)

PR #360, opened 2016-12-06
Wday weekly schedule problem

MGPalmer:
Hello!
This fixes a bug that occurs when using weekly schedules and having monday as the week start. On some days, the bug would cause occurrences to be missed.
The fix is simply using the wday normalizing method that is being used everywhere else, as far as I can see it has been overlooked. Also, lots of collected tests are added.

### DIFF
--- a/lib/ice_cube/schedule.rb
+++ b/lib/ice_cube/schedule.rb
@@ -510,21 +510,18 @@ module IceCube
     # earlier validations in the week). This realigns the opening time to
     # the start of the interval's correct period (e.g. move to start of week N)
     #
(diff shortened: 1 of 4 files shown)


avit:
Thanks for this. I believe your implementation is missing a conditional otherwise it would cause errors on rules that don't respond to `week_start`. I ended up moving the logic into the Rule classes instead of adding more logic into Schedule. See #383.

MGPalmer:
Ohai! Coming back to this after a long time to say thank you :) We've finally found time to update and go back to your current master and this really helps!

PR #369, opened 2017-01-12
Fix tests on older rubies

seejohnrun:
Don't try to bring in new activesupport on versions of Ruby that don't
support it

### DIFF
--- a/ice_cube.gemspec
+++ b/ice_cube.gemspec
@@ -19,7 +19,7 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency('rake')
   s.add_development_dependency('rspec', '~> 2.12.0')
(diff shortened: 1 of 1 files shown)


hwhelchel:
@seejohnrun thanks for taking a crack at this. I agree with your use of `Gem::Version`. After I submitted my PR here, I fixed another 2.4 issue on another gem [here](https://github.com/plasticine/middleman-imageoptim/pull/48/files#diff-246e7a249153e33a54f03fe115f857dcR91) using that api.

Here is how Rails handled supporting 2.4 in rails 4.2: https://github.com/rails/rails/pull/27473

seejohnrun:
my pleasure
I'm merging this for now, can you rebase your branch on top of this?

PR #373, opened 2017-01-27
Bugs: Fix more invalid rules

Dorian:
- [x] Fix specs

(A fuzzer could be better at this than me :D )

### DIFF
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -31,6 +31,7 @@ module IceCube
 
       ical.split(';').each do |rule|
         (name, value) = rule.split('=')
+        raise ArgumentError, "Invalid iCal rule component" if value.nil?
         value.strip!
         case name
         when 'FREQ'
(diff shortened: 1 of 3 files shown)


lib/ice_cube/rule.rb
```diff
@@ -73,6 +73,7 @@ class << self
       # Convert from a hash and create a rule
       def from_hash(original_hash)
         hash = IceCube::FlexibleHash.new original_hash
+        raise ArgumentError, 'empty ical rule_type' if hash[:rule_type].nil?
```
seejohnrun: I think we can do something more here by saying:

`raise ArgumentError, 'Invalid rule type' unless hash[:rule_type] && hash[:rule_type].match(/\:\:(.+?)Rule/)`

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -31,6 +31,7 @@ def self.rule_from_ical(ical)
 
       ical.split(';').each do |rule|
         (name, value) = rule.split('=')
+        raise ArgumentError, "empty ical rule's value" if value.nil?
```
seejohnrun: I think the text of this error should be something more like: "Invalid iCal rule component"

Dorian:
@seejohnrun What about now?

seejohnrun:
👍

PR #377, opened 2017-02-23
Fixing biweekly realign when used with spans option.

jopotts:
This PR fixes an edge case bug for the use of the spans option when the biweekly realign method comes into effect. The realign method was being called with the opening time with the duration shift, whereas it should always be using the pure opening time.

Thanks for all the work you've been doing on the gem recently! It's not gone unnoticed.


seejohnrun:
Thanks @jopotts and thank you for the note :)

PR #379, opened 2017-03-19
Use Integer class to avoid Fixnum deprecation warnings in ruby 2.4

avit:
This looks good. Thanks @patrickjaberg!


loed-idzinga:
Thank you both for the hard work! Is it possible to make a release of this fix? 👍

PR #383, opened 2017-03-20
Fix first weekly occurrences with non-Sunday week start

avit:
This change corrects the wday offset to account for the week start option in weekly rules.

Some refactoring was done to avoid adding conditionals and more complexity into the Schedule class which was coordinating this:

- Rules no longer depend on a Schedule object to validate: to simplify, they only require Time objects and calculate the next time coordinate from them.
- Rules no longer have a second argument for `week_start` except for the Weekly rule that uses it. This is to avoid confusion about where this option is actually used.

Much thanks to #360 and #284 for spotting the core issue here.

### DIFF
--- a/lib/ice_cube/rules/hourly_rule.rb
+++ b/lib/ice_cube/rules/hourly_rule.rb
@@ -4,7 +4,7 @@ module IceCube
 
     include Validations::HourlyInterval
 
-    def initialize(interval = 1, week_start = :sunday)
+    def initialize(interval = 1)
       super
       interval(interval)
       schedule_lock(:min, :sec)
--- a/lib/ice_cube/rules/weekly_rule.rb
+++ b/lib/ice_cube/rules/weekly_rule.rb
@@ -13,16 +13,28 @@ module IceCube
       reset
     end
 
+    # Calculate the effective start time for when the given start time is later
+    # in the week than one of the weekday validations, such that times could be
+    # missed by a 7-day jump using the weekly interval, or when selecting from a
+    # date that is misaligned from the schedule interval.
+    #
+    def realign(step_time, start_time)
+      time = TimeUtil::TimeWrapper.new(start_time)
+      offset = wday_offset(step_time, start_time)
+      time.add(:day, offset) if offset
+      time.to_time
+    end
+
     def wday_offset(step_time, start_time)
       wday_validations = other_interval_validations.select { |v| v.type == :wday }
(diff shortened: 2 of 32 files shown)


seejohnrun:
Sorry for the delay - will have this back to you tonight!

avit:
Thanks, I'd just like a second pair of eyes on this since it's pretty big. (I hope it should make sense to review per-commit.)

lib/ice_cube/validations/hourly_interval.rb
```diff
@@ -25,8 +25,8 @@ def dst_adjust?
         false
       end
 
-      def validate(step_time, schedule)
-        t0, t1 = schedule.start_time.to_i, step_time.to_i
+      def validate(step_time, start_time)
```
seejohnrun: 👍

lib/ice_cube/rules/hourly_rule.rb
```diff
@@ -4,7 +4,7 @@ class HourlyRule < ValidatedRule
 
     include Validations::HourlyInterval
 
-    def initialize(interval = 1, week_start = :sunday)
+    def initialize(interval = 1)
```
seejohnrun: always thought these were a little clunky so I'm happy to see them go

lib/ice_cube/rules/weekly_rule.rb
```diff
@@ -4,13 +4,39 @@ class WeeklyRule < ValidatedRule
...
+      time.to_time
+    end
+
+    def wday_offset(step_time, start_time)
+      wday_validations = other_interval_validations.select { |v| v.type == :wday }
+      return if wday_validations.none?
+
+      days = (step_time - start_time).to_i / ONE_DAY
+      interval = base_interval_validation.validate(step_time, start_time).to_i
+      min_wday = TimeUtil.normalize_wday(wday_validations.min_by(&:day).day, week_start)
```
seejohnrun: I believe here this is the equivalent of `map(&:day).min` which may read nicer

avit: Did it this way because `min_by` takes one pass over the array and `map.min` takes two.

(However, this method is only called once per query, and it's a tiny array of 7 so it's not going to be significant either way.)

seejohnrun: Definitely - sounds good!

PR #394, opened 2017-05-05
[WIP] Fix #377 regression

jopotts:
The recent refactor of the weekly rule realignment has raised an error. This PR adds tests to show the problem and an attempt to fix it (currently not working).

The problem is that the duration offset used to be applied to the initial time t1 in the enumerator after the realignment was applied, but it has now been pushed back to the rule, so the realignment was being done on the offset t1 rather than the original opening_time. See #377

### DIFF
--- a/lib/ice_cube/schedule.rb
+++ b/lib/ice_cube/schedule.rb
@@ -426,16 +426,17 @@ module IceCube
       opening_time += start_time.subsec - opening_time.subsec rescue 0
       opening_time = start_time if opening_time < start_time
       spans = options[:spans] == true && duration != 0
+      spans_offset = spans ? duration : 0
       Enumerator.new do |yielder|
         reset
-        t1 = full_required? ? start_time : opening_time - (spans ? duration : 0)
+        t1 = full_required? ? start_time : opening_time
         loop do
-          break unless (t0 = next_time(t1, closing_time))
+          break unless (t0 = next_time(t1, closing_time, spans_offset))
           break if closing_time && t0 > closing_time
           if (spans ? (t0.end_time > opening_time) : (t0 >= opening_time))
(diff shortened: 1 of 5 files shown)


jopotts:
Don't merge as the tests are currently failing. I have to admit that I'm a bit stumped on the best way to fix this, so it would be super appreciated if you could have a look. Thanks!

lib/ice_cube/schedule.rb
```diff
@@ -426,16 +426,17 @@ def enumerate_occurrences(opening_time, closing_time = nil, options = {}, &block
       opening_time += start_time.subsec - opening_time.subsec rescue 0
       opening_time = start_time if opening_time < start_time
       spans = options[:spans] == true && duration != 0
+      spans_offset = spans ? duration : 0
       Enumerator.new do |yielder|
         reset
-        t1 = full_required? ? start_time : opening_time - (spans ? duration : 0)
```
jopotts: `realign` used to happen here before the spans offset was applied.

lib/ice_cube/validated_rule.rb
```diff
@@ -57,11 +57,12 @@ def base_interval_type
 
     # Compute the next time after (or including) the specified time in respect
     # to the given start time
-    def next_time(time, start_time, closing_time)
+    def next_time(time, start_time, closing_time, spans_offset = 0)
       @time = time
       unless @start_time
         @start_time = realign(time, start_time)
         @time = @start_time if @time < @start_time
+        @time = @time - spans_offset
```
jopotts: Spans offset now applied here after realignment, but some tests still fail.

avit:
Thanks, I'll try and dig into this one soon.

Issue #400, opened 2017-06-27
Missing occurrences on weekly schedules with weekstart mondays


MGPalmer:
If you add a schedule like this:
* week start set to monday
* start time on a monday
* weekly
* every day of the week except tuesday

then it will miss the first occurrences...I'll attach a pull request with a spec inn a minute.

MGPalmer:
Ah, update: Yes, this bug only happens with week start monday...unfortunately, this is what our part of the world uses :/

MGPalmer:
Update II: I think I fixed the original problem (the spec I added is green). However, in our production data I found another issue (Biweekly schedules occurring mondays and thursdays, with monday weekstart, seen from a specific date), which I can't fix. I added a spec for this case. All is in the PR.

MGPalmer:
@seejohnrun @avit Sooo, I've been pulling my hair out over this issue - it seems to me like there is a general problem with weekly schedules and starting the week on monday, but I can't grok the code enough to find what's going on. This is a quite bad problem for our service, we keep getting reports about missing occurrences.
Is there any chance of anyone taking a look at this? We'd be willing to offer some sort of bounty for getting timely help :)

seejohnrun:
@MGPalmer I may be able to look tonight or tomorrow morning - email at your convenience. first.last@gmail.com

MGPalmer:
@seejohnrun You've got mail :)

seejohnrun:
@MGPalmer haven't received anything - make sure you put my first and last names in there 😉

MGPalmer:
Strange, I took it from your website...I'll send again from a different account.

PR #401, opened 2017-06-27
Add spec illustrating a probable bug - for ticket #400

MGPalmer:
#400


avit:
Fixed, I used your tests as a template, thanks!

PR #403, opened 2017-06-29
Fix weekly realignment for `spans: true` option

avit:
Clamp iteration to the schedule start time so the first iteration does not
appear to be a preceding day (or possibly week).

When using the `spans: true` option, reversing over the start time to cover the
entire duration can move the opening time into a previous day/week and cause
weekly realignment to jump into the next weekly interval.

Count whole days for realigning weekly offsets to prevent rounding down to the
previous day.

Fixes #394


jopotts:
Amazing! Only just spotted this now. Thank you!

PR #408, opened 2017-08-04
Full Timezone support when parsing ical

seejohnrun:
This PR builds on the great work from #335 (thanks @dcosson)

I'll be removing the need for `ActiveSupport` here.

Also, added support for parsing `RDATE` while I'm in here.

### DIFF
--- a/ice_cube.gemspec
+++ b/ice_cube.gemspec
@@ -22,4 +22,5 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency('rake')
   s.add_development_dependency('rspec', '> 3')
+  s.add_development_dependency('timecop')
 end
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -1,83 +1,102 @@
 module IceCube
   class IcalParser
-    def self.schedule_from_ical(ical_string, options = {})
-      data = {}
-      ical_string.each_line do |line|
-        (property, value) = line.split(':')
-        (property, tzid) = property.split(';')
-        case property
-        when 'DTSTART'
-          data[:start_time] = Time.parse(value)
-        when 'DTEND'
-          data[:end_time] = Time.parse(value)
-        when 'EXDATE'
-          data[:extimes] ||= []
-          data[:extimes] += value.split(',').map{|v| Time.parse(v)}
-        when 'DURATION'
-          data[:duration] # FIXME
-        when 'RRULE'
-          data[:rrules] ||= []
-          data[:rrules] += [rule_from_ical(value)]
-        end
-      end
-      Schedule.from_hash data
-    end
-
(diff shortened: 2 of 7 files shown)


seejohnrun:
@avit do you think we should move ical parsing out of `ice_cube` proper as part of this?  Then we could avoid having to wrap with `active_support` conditionals.

ice_cube.gemspec
```diff
@@ -22,4 +22,5 @@ Gem::Specification.new do |s|
 
   s.add_development_dependency('rake')
   s.add_development_dependency('rspec', '> 3')
+  s.add_development_dependency('timecop')
```
avit: Just curious, what do you use it for? I always just stub `Time.now` and never found a need for this gem! 😉

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -1,83 +1,102 @@
...
+        Rule.from_hash(params)
       end
 
-      Rule.from_hash(params)
+      private
+
+      def parse_in_tzid(value, tzid)
+        time_parser = Time
+        if tzid
+          time_parser = ActiveSupport::TimeZone.new(tzid.split('=')[1]) || Time
```
avit: Use `rescue Time` here if ActiveSupport is not loaded? Maybe something like `IcalParser.time_parser` can be determined statically for performance.

avit:
I like the idea... are you thinking of a separate gem, or maybe just an optional require?

seejohnrun:
I think the optional require makes sense, but separate gem also sounds like a decent idea to me.  I just dislike that we're potentially having two different behaviors in the ical parser dependending if `AS` is loaded or not

avit:
We have a dependency on ActiveSupport [here][1]. Maybe this can work with `TimeUtil.deserialize_time` by passing a hash to it.


[1]: https://github.com/seejohnrun/ice_cube/blob/master/lib/ice_cube/time_util.rb#L115

seejohnrun:
@avit Ah I like that idea - will update this soon

dcosson:
Any update on this?

It's really a bummer that this has been roadblocked for years at this point, even though the current behavior which is completely broken w.r.t. daylight savings time, just to avoid the dependency on ActiveSupport. With all due respect, what's more important - having a library that works and solves people's problems, or keeping the dependencies down?

seejohnrun:
What do people think of the compromise of just making this method raise if `ActiveSupport` hasn't been required and we're attempting to parse a time with zone?  I'm not 100% against the idea of the added dependency anymore, but want to make sure we're doing it for a reason that is required.  I'm sorry this has held you up @dcosson and hopefully this branch has been useful to you.

dcosson:
No worries, thanks for your continuing work maintaining this library. Hopefully it's close now!

+1 from me on raising if ActiveSupport isn't available.

thefliik:
Seems like a good idea.

PR #420, opened 2017-09-18
Prevent illogical rule combinations

### LINKED ISSUE #406: Weekly rule when first occurrence is the next day and hour before current hour
Version 0.16.2 seems to have a problem when creating a weekly schedule when the first occurrence happens to be the next day and the hour_of_day is before the current hour of day.  When this happens, the next_occurrence is pushed out an extra week.  If the hour_of_day is equal to or after the current hour of day, it works fine.

This test works with 0.16.0 but not with 0.16.2
  

    test 'weekly_schedule_flow_next_day_before_hour_test' do

      schedule = IceCube::Schedule.new(Time.utc(2017, 7, 28, 20, 0, 0))

      schedule.add_recurrence_rule IceCube::Rule.weekly(1).day(:saturday).hour_of_day(19)

      result = schedule.next_occurrence(Time.utc(2017, 7, 28, 20, 0, 0))

      assert_equal Time.utc(2017, 7, 29, 19, 0, 0), result, "Should be #{Time.utc(2017, 7, 29, 19, 0, 0)} but got #{result}"

    end

Minitest::Assertion: Should be 2017-07-29 19:00:00 UTC but got 2017-08-05 19:00:00 UTC.

Expected: 2017-07-29 19:00:00 UTC
Actual: 2017-08-05 19:00:00 UTC

### LINKED ISSUE #416: Rule validations that don't align with the start time run indefinitely
For example:

```
Rule.minutely(30).minute_of_hour(15, 45)
```

In order to find any matching times, this validation needs to align with the minute on the schedule start time.

⚠️ If the start time is misaligned then it **loops indefinitely** looking for the next valid time.

Some of these combinations are certainly invalid. When a rule (of any frequency) is set to recur by some interval, then validating occurrences of the same frequency may be invalid, even though IceCube accepts them.

Moving up to the next frequency seems more correct and works as expected:

```
Rule.hourly.minute_of_hour(15, 45)
```

While this "Hourly on the 15th and 45th minutes of the hour" schedule doesn't read as naturally as "Every 30 minutes on the 15th and 45th minutes of the hour", it does avoid near-impossible combinations such as:

```
Rule.minutely(10).minute_of_hour(3, 6, 9)
```

It's not even logical what should happen with something like "Every 10 minutes on the 3rd, 6th, and 9th minutes of the hour". After the obvious first occurrence at minute :03 should we:

1. Advance to next minute-of-hour (minute :06)
2. Advance to the next interval (minute :13)

I would argue that this is invalid input in any case. The RFC [specifies][1] some "rule parts" as invalid on all recurrence rules, but does not prevent some of the problematic combinations, which I've highlighted here:

| |SECONDLY|MINUTELY|HOURLY|DAILY|WEEKLY|MONTHLY|YEARLY|
|---|---|---|---|---|---|---|---|
|BYMONTH|Limit|Limit|Limit|Limit|Limit |Limit ⚠️|Expand|
|BYWEEKNO|N/A|N/A|N/A|N/A|N/A|N/A|Expand|
|BYYEARDAY|Limit|Limit|Limit|N/A|N/A|N/A|Expand|
|BYMONTHDAY|Limit|Limit|Limit|Limit ⚠️|N/A|Expand |Expand|
|BYDAY|Limit|Limit|Limit|Limit ⚠️|Expand|Note 1 |Note 2|
|BYHOUR|Limit|Limit|Limit ⚠️|Expand |Expand|Expand|Expand|
|BYMINUTE|Limit|Limit ⚠️|Expand|Expand |Expand|Expand |Expand|
|BYSECOND|Limit ⚠️|Expand|Expand|Expand|Expand|Expand|Expand|
|BYSETPOS|Limit|Limit|Limit|Limit|Limit|Limit|Limit|

[1]: https://tools.ietf.org/html/rfc5545#section-3.3.10

We can easily remove access to some of these "N/A" combinations, but nothing is specified for handling the conflicted ones. Here's what we can do:

### When the interval is 1

```
Rule.hourly.hour_of_day(3,6,9)
```

This does not conflict since every hour is evaluated to see if it matches, and is thus _limited_ to the given hours.

### When the interval is > 1

We could realign the schedule start time to the validated time part. Example:

```
Rule.hourly(3).hour_of_day(3)
```

This could realign the schedule start time to hour = 3, and would effectively return daily occurrences only since it limits to the given hour only.

### When there are multiple values

This is when things can get strange, since the additional values don't always align:

```
Rule.hourly(5).hour_of_day(5, 10, 15)
```

Assuming we realign to the first hour (hour = 5) as above, then the first day could have 3 occurrences, and the following 4 days would have none (the 5 hour interval means 25 hours to the next day and the hours would misalign as [1, 6, 11, 16]; then [2, 7, 12, 17]...

If the interval and validations aren't even multiples like above, then it might only realign itself very rarely. One way to deal with this would be to reject anything that is not a multiple, e.g. "every 30 minutes on the 15th and 45th minute" could be easy considered as correct since difference of 45 – 15 matches the 30 minute interval. I can't imagine any logical situations where allowing non-multiples would make any sense, and these should raise an argument error.

avit:
Fixes #416 
Fixes #406

### DIFF
--- a/lib/ice_cube/rules/daily_rule.rb
+++ b/lib/ice_cube/rules/daily_rule.rb
@@ -11,6 +11,26 @@ module IceCube
       reset
     end
 
+    def verify_alignment(value, freq, rule_part)
+      return unless freq == :wday || freq == :day
+      return unless @validations[:interval]
+
+      interval_validation = @validations[:interval].first
+      interval_value = (rule_part == :interval) ? value : interval_validation.interval
+      return if interval_value == 1
+
+      if freq == :wday
+        return if (interval_value % 7).zero?
+        return if Array(@validations[:day]).empty?
+        message = "day can only be used with multiples of interval(7)"
+      else
+        (fixed_validation = other_fixed_value_validations.first) or return
+        message = "#{fixed_validation.key} can only be used with interval(1)"
+      end
+
+      yield ArgumentError.new(message)
+    end
+
   end
 
 end
--- a/lib/ice_cube/rules/daily_rule.rb
+++ b/lib/ice_cube/rules/daily_rule.rb
@@ -2,6 +2,15 @@ module IceCube
 
   class DailyRule < ValidatedRule
 
+    include Validations::HourOfDay
+    include Validations::MinuteOfHour
+    include Validations::SecondOfMinute
+    include Validations::DayOfMonth
+    include Validations::DayOfWeek
+    include Validations::Day
+    include Validations::MonthOfYear
+    # include Validations::DayOfYear    # n/a
+
     include Validations::DailyInterval
 
     def initialize(interval = 1)
--- a/lib/ice_cube/rules/daily_rule.rb
+++ b/lib/ice_cube/rules/daily_rule.rb
@@ -20,26 +20,6 @@ module IceCube
       reset
     end
 
-    def verify_alignment(value, freq, rule_part)
-      return unless freq == :wday || freq == :day
-      return unless @validations[:interval]
-
-      interval_validation = @validations[:interval].first
-      interval_value = (rule_part == :interval) ? value : interval_validation.interval
-      return if interval_value == 1
-
-      if freq == :wday
-        return if (interval_value % 7).zero?
-        return if Array(@validations[:day]).empty?
-        message = "day can only be used with multiples of interval(7)"
-      else
-        (fixed_validation = other_fixed_value_validations.first) or return
-        message = "#{fixed_validation.key} can only be used with interval(1)"
-      end
-
-      yield ArgumentError.new(message)
-    end
-
   end
 
 end
--- a/lib/ice_cube/validated_rule.rb
+++ b/lib/ice_cube/validated_rule.rb
@@ -51,6 +51,14 @@ module IceCube
       Array(@validations[base_interval_validation.type])
     end
 
+    def other_fixed_value_validations
+      @validations.values.flatten.select { |v|
+        interval_type = (v.type == :wday ? :day : v.type)
+        v.class < Validations::FixedValue &&
+          interval_type == base_interval_validation.type
+      }
+    end
+
     # Compute the next time after (or including) the specified time in respect
     # to the given start time
     def next_time(time, start_time, closing_time)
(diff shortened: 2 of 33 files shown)


spec/examples/regression_spec.rb
```diff
@@ -36,12 +36,6 @@ module IceCube
           expect(schedule.occurrences(Date.today >> 12)).to be_an Array
         end
 
-        it 'should not regress [#40]' do
-          schedule = Schedule.new(Time.local(2011, 11, 16, 11, 31, 58), :duration => 3600)
-          schedule.add_recurrence_rule Rule.minutely(60).day(4).hour_of_day(14, 15, 16).minute_of_hour(0)
-          expect(schedule.occurring_at?(Time.local(2011, 11, 17, 15, 30))).to be_falsey
-        end
-
```
avit: This spec originally fixed an infinite loop by skipping over valid occurrences. It should actually occur on the time in this example so I removed it.

lib/ice_cube/validated_rule.rb
```diff
@@ -4,15 +4,6 @@ class ValidatedRule < Rule
 
     include Validations::ScheduleLock
 
-    include Validations::HourOfDay
```
seejohnrun: 👍 to splitting these out so we can only have them where they make sense

lib/ice_cube/rules/daily_rule.rb
```diff
@@ -11,6 +20,26 @@ def initialize(interval = 1)
       reset
     end
 
+    def verify_alignment(value, freq, rule_part)
+      return unless freq == :wday || freq == :day
+      return unless @validations[:interval]
+
+      interval_validation = @validations[:interval].first
```
seejohnrun: Maybe worth moving this interval-checking into a util that we can use in both places

lib/ice_cube/validated_rule.rb
```diff
@@ -51,6 +42,14 @@ def other_interval_validations
       Array(@validations[base_interval_validation.type])
     end
 
+    def other_fixed_value_validations
+      @validations.values.flatten.select { |v|
```
seejohnrun: nit: `do / end` on multi-line blocks

lib/ice_cube/validations/daily_interval.rb
```diff
@@ -4,7 +4,11 @@ module Validations::DailyInterval
 
     # Add a new interval validation
     def interval(interval)
-      @interval = normalized_interval(interval)
+      interval = normalized_interval(interval)
+      verify_alignment(interval, :wday, :interval) { |error| raise error }
```
seejohnrun: Why do we yield this error up instead of just `raise`ing it?

avit: This is so that it gives a more relevant trace, since it appears from the called `interval` public method interface, instead of raising from somewhere deeper down.

lib/ice_cube/rules/daily_rule.rb
avit: Yes, I'm not 100% happy with some of the duplication & super-calling either. I'll see what I can do.

lib/ice_cube/validated_rule.rb
avit: Ok, sure. 😄 

I've always been more partial to [Jim Weirich's semantic brace style][1], and I try to sneak it in when I can:

- Use { } for blocks that return values (functional)
- Use do / end for blocks that are executed for side effects (procedural)

It actually turns out that doing it this way, it works out that _very often_ it's the same as the single-line vs. multi-line style, because of the kinds of things that go into procedural vs. functional blocks.

[1]: http://archive.is/96ko8

seejohnrun: Okay I can be down with that - I like the idea :)

PR #428, opened 2017-10-16
Fix Schedule#exception_rules

### LINKED ISSUE #427: Schedule#exception_rules behavior is incorrect
it should be:
```
    # Get the exception rules
    def exception_rules
      @all_exception_rules.select { |r| r.is_a?(SingleOccurrenceRule) }
    end
```

Though it would be better if there is an `ExceptionRule` class or something alike.

dsdshcym:
- fixes #427
- since rules in @all_exception_rules are all SingleOccurrenceRule, so
  it doesn't make any sense to reject SingleOccurrenceRule objects

### DIFF
--- a/lib/ice_cube/schedule.rb
+++ b/lib/ice_cube/schedule.rb
@@ -100,7 +100,7 @@ module IceCube
 
     # Get the exception rules
     def exception_rules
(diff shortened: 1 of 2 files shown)


seejohnrun:
I believe this is because single occurrence rules are actually `exception_times`, not `exception_rules`.

avit:
Closing, see discussion on linked issue.

Issue #431, opened 2017-12-18
Translations for date overriding main project translations


MGPalmer:
Hi!
I've been lagging behind the ice_cube master for a while now, and while checking out the last changes, I noticed some of our date/time display was off. I tracked it down to one commit, I think: https://github.com/seejohnrun/ice_cube/commit/c7b305c3f1abc98102dd28e473ecd6983b8d3bf8

Let me try to illustrate:
This is a Rails app with the ice_cube gem loaded via a Gemfile, but with a local bundler override so it uses my local repo of ice_cube. With my very old master (at d8614f739bc6edf51fb5fb18c9fd19f710c2d9da) , the default date translation looks like this:

```
cormorant:~/Development/promptus/sports-mssngr [master] $ rails c
Loading development environment (Rails 4.2.10)
mssngr(dev)> I18n.t("date.formats.default", locale: :de)
=> "%d.%m.%Y"
mssngr(dev)> IceCube::Schedule.new.to_s
=> "December 18, 2017"
mssngr(dev)> I18n.t("date.formats.default", locale: :de)
=> "%d.%m.%Y"
```

but after switching to upstream:

```
cormorant:~/Development/promptus/ice_cube [master] $ git co c7b305c3f1abc98102dd28e473ecd6983b8d3bf8
Note: checking out 'c7b305c3f1abc98102dd28e473ecd6983b8d3bf8'.
HEAD is now at c7b305c... Fix core extension load order between ActiveSupport and I18n
```
...loading the ice cube code now clobbers our translation:

```
cormorant:~/Development/promptus/sports-mssngr [master] $ rails c
Loading development environment (Rails 4.2.10)
mssngr(dev)> I18n.t("date.formats.default", locale: :de)
=> "%d.%m.%Y"
mssngr(dev)> IceCube::Schedule.new.to_s
=> "December 18, 2017"
mssngr(dev)> I18n.t("date.formats.default", locale: :de)
=> "%Y-%m-%d"

```

(btw, "%Y-%m-%d" is also not really used in germany ;))

seejohnrun:
@MGPalmer what do you think of submitting a PR to correct the date format? cc / @joelmeyerhamme

MGPalmer:
@seejohnrun Sure I can do that, but that wouldn't fix the translation-clobbering issue, right?

MGPalmer:
[...]

PR #432, opened 2017-12-19
Remove "overriding" translation keys

MGPalmer:
Refs #431

All specs still pass, and this solves my overriding problem :)

### DIFF
--- a/config/locales/de.yml
+++ b/config/locales/de.yml
@@ -146,10 +146,6 @@ de:
     - Donnerstag
     - Freitag
     - Samstag
-    formats:
-      default: "%Y-%m-%d"
-      long: "%B %d, %Y"
-      short: "%b %d"
     month_names:
     -
     - Januar
@@ -168,11 +164,3 @@ de:
     - :year
     - :month
     - :day
-
-  time:
-    am: am
-    formats:
-      default: "%a, %d %b %Y %H:%M:%S %z"
-      long: "%B %d, %Y %H:%M"
-      short: "%d %b %H:%M"
-    pm: pm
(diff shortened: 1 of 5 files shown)


seejohnrun:
@MGPalmer thanks! do you mind adding a test or two to make sure formats can differ between locales? I'm unsure if they already exist

MGPalmer:
Hmm I'm not 100% sure I understand what you'd like me to test...

However, looking over the code and specs, it seems to me like only the ice_cube....formats.default is being used, and that is pretty well tested.

In fact, setting a different to_s format seems broken:

```
mssngr(dev)> IceCube.to_s_time_format = "aaaa"
=> "aaaa"
mssngr(dev)> IceCube.to_s_time_format
=> "%B %-d, %Y"

```

Maybe because the instance variable set is not actually used in this getter:

```
  # Defines the format used by IceCube when printing out Schedule#to_s.
  # Defaults to '%B %e, %Y'
  def self.to_s_time_format
    IceCube::I18n.t("ice_cube.date.formats.default")
  end

  # Sets the format used by IceCube when printing out Schedule#to_s.
  def self.to_s_time_format=(format)
    @to_s_time_format = format
  end

```

but that's a different ticket ;)

avit:
> but that's a different ticket ;)

Yes, it looks like that setting was just left behind after translations were added. We should remove since it does nothing, but that could break code which actually sets it, so it should have a deprecation first.

avit:
Also I'm surprised why the test suite is all green here with these removed... something to look into.

MGPalmer:
> Also I'm surprised why the test suite is all green here with these removed... something to look into.

I might be mistaken, but my guess is that the format keys I've removed have not actually been used in ice_cube.

> The real problem is that previously, the load order was causing other problems and now it sounds like we are overwriting these fields the wrong way. I will take another look at the c7b305c commit which broke this.

This would of course be my favorite solution!

PR #449, opened 2018-07-30
Update to BYSETPOS support

nehresma:
In August 2016, Nicolas Marlier added BYSETPOS support (#349).  This PR updates that PR with a few small changes to run in modern Ruby (use Integer instead of Fixnum) and a more modern rspec.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,8 +8,5 @@
 /spec/reports/
 /tmp/
 
-# rubymine
-.idea
-
 # rspec failure tracking
 .rspec_status
--- a/.gitignore
+++ b/.gitignore
@@ -8,5 +8,8 @@
 /spec/reports/
 /tmp/
 
+# rubymine
+.idea
+
 # rspec failure tracking
 .rspec_status
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -6,7 +6,7 @@ module IceCube
       # First join lines that are wrapped
       lines = []
       ical_string.each_line do |line|
-        if lines[-1] && line =~ /\A[ \t].+\z/
+        if lines[-1] && line =~ /\A[ \t].+/
           lines[-1] = lines[-1].strip + line.sub(/\A[ \t]+/, "")
         else
           lines << line
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -2,7 +2,18 @@ module IceCube
   class IcalParser
     def self.schedule_from_ical(ical_string, options = {})
       data = {}
+
+      # First join lines that are wrapped
+      lines = []
       ical_string.each_line do |line|
+        if lines[-1] && line =~ /\A[ \t].+\z/
+          lines[-1] = lines[-1].strip + line.sub(/\A[ \t]+/, "")
+        else
+          lines << line
+        end
+      end
+
+      lines.each do |line|
         (property, value) = line.split(":")
         (property, _tzid) = property.split(";")
         case property
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -75,7 +75,7 @@ module IceCube
         when "BYYEARDAY"
           validations[:day_of_year] = value.split(",").map(&:to_i)
         when "BYSETPOS"
-          # noop
+          params[:validations][:by_set_pos] = value.split(',').collect(&:to_i)
         else
           validations[name] = nil # invalid type
         end
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -86,7 +86,7 @@ module IceCube
         when "BYYEARDAY"
           validations[:day_of_year] = value.split(",").map(&:to_i)
         when "BYSETPOS"
-          params[:validations][:by_set_pos] = value.split(',').collect(&:to_i)
+          validations[:by_set_pos] = value.split(",").map(&:to_i)
         else
           validations[name] = nil # invalid type
         end
--- a/lib/ice_cube/time_util.rb
+++ b/lib/ice_cube/time_util.rb
@@ -1,7 +1,5 @@
 require 'date'
 require 'time'
-require 'active_support'
-require 'active_support/core_ext'
 
 module IceCube
   module TimeUtil
--- a/lib/ice_cube/time_util.rb
+++ b/lib/ice_cube/time_util.rb
@@ -195,36 +195,6 @@ module IceCube
       [nth_occurrence_of_weekday, this_weekday_in_month_count]
     end
 
-    # Use Activesupport CoreExt functions to manipulate time
-    def self.start_of_month time
-      time.beginning_of_month
-    end
-
-    # Use Activesupport CoreExt functions to manipulate time
-    def self.end_of_month time
-      time.end_of_month
-    end
-
-    # Use Activesupport CoreExt functions to manipulate time
-    def self.start_of_year time
-      time.beginning_of_year
-    end
-
-    # Use Activesupport CoreExt functions to manipulate time
-    def self.end_of_year time
-      time.end_of_year
-    end
-
-    # Use Activesupport CoreExt functions to manipulate time
-    def self.previous_month time
-      time - 1.month
-    end
-
-    # Use Activesupport CoreExt functions to manipulate time
-    def self.previous_year time
-      time - 1.year
-    end
-
     # Get the days in the month for +time
     def self.days_in_month(time)
       date = Date.new(time.year, time.month, 1)
--- a/lib/ice_cube/time_util.rb
+++ b/lib/ice_cube/time_util.rb
@@ -1,5 +1,7 @@
-require "date"
-require "time"
+require 'date'
+require 'time'
+require 'active_support'
+require 'active_support/core_ext'
 
 module IceCube
   module TimeUtil
@@ -191,6 +193,36 @@ module IceCube
       this_weekday_in_month_count = ((days_in_month(time) - first_occurrence + 1) / 7.0).ceil
       nth_occurrence_of_weekday = (time.mday - first_occurrence) / 7 + 1
       [nth_occurrence_of_weekday, this_weekday_in_month_count]
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.start_of_month time
+      time.beginning_of_month
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.end_of_month time
+      time.end_of_month
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.start_of_year time
+      time.beginning_of_year
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.end_of_year time
+      time.end_of_year
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.previous_month time
+      time - 1.month
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.previous_year time
+      time - 1.year
     end
 
     # Get the days in the month for +time
--- a/lib/ice_cube/time_util.rb
+++ b/lib/ice_cube/time_util.rb
@@ -1,5 +1,5 @@
-require 'date'
-require 'time'
+require "date"
+require "time"
 
 module IceCube
   module TimeUtil
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -1,0 +1,87 @@
+module IceCube
+
+  module Validations::MonthlyBySetPos
+
+    def by_set_pos(*by_set_pos)
+      return by_set_pos([by_set_pos]) if by_set_pos.is_a?(Fixnum)
+
+      unless by_set_pos.nil? || by_set_pos.is_a?(Array)
+        raise ArgumentError, "Expecting Array or nil value for count, got #{by_set_pos.inspect}"
+      end
+      by_set_pos.flatten!
+      by_set_pos.each do |set_pos|
+        unless (set_pos >= -366 && set_pos <= -1) ||
+            (set_pos <= 366 && set_pos >= 1)
+          raise ArgumentError, "Expecting number in [-366, -1] or [1, 366], got #{set_pos} (#{by_set_pos})"
+        end
+      end
+
+      @by_set_pos = by_set_pos
+      replace_validations_for(:by_set_pos, by_set_pos && [Validation.new(by_set_pos, self)])
+      self
+    end
+
+    class Validation
+
+      attr_reader :rule, :by_set_pos
+
+      def initialize(by_set_pos, rule)
+
+        @by_set_pos = by_set_pos
+        @rule = rule
+      end
+
+      def type
+        :day
+      end
+
+      def dst_adjust?
+        true
+      end
+
+      def validate(step_time, schedule)
+        start_of_month = TimeUtil.start_of_month step_time
+        end_of_month = TimeUtil.end_of_month step_time
+
+
+        new_schedule = IceCube::Schedule.new(TimeUtil.previous_month(step_time)) do |s|
+          s.add_recurrence_rule IceCube::Rule.from_hash(rule.to_hash.reject{|k, v| [:by_set_pos, :count, :until].include? k})
+        end
+
+        puts step_time
+        occurrences = new_schedule.occurrences_between(start_of_month, end_of_month)
+        p occurrences
+        index = occurrences.index(step_time)
+        if index == nil
+          1
+        else
+          positive_set_pos = index + 1
+          negative_set_pos = index - occurrences.length
+
+          if @by_set_pos.include?(positive_set_pos) || @by_set_pos.include?(negative_set_pos)
+            0
+          else
+            1
+          end
+        end
+      end
+
+
+      def build_s(builder)
+        builder.piece(:by_set_pos) << by_set_pos
+      end
+
+      def build_hash(builder)
+        builder[:by_set_pos] = by_set_pos
+      end
+
+      def build_ical(builder)
+        builder['BYSETPOS'] << by_set_pos
+      end
+
+      nil
+    end
+
+  end
+
+end
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -1,23 +1,16 @@
 module IceCube
 
   module Validations::MonthlyBySetPos
-
     def by_set_pos(*by_set_pos)
-      return by_set_pos([by_set_pos]) if by_set_pos.is_a?(Integer)
-
-      unless by_set_pos.nil? || by_set_pos.is_a?(Array)
-        raise ArgumentError, "Expecting Array or nil value for count, got #{by_set_pos.inspect}"
-      end
       by_set_pos.flatten!
       by_set_pos.each do |set_pos|
-        unless (set_pos >= -366 && set_pos <= -1) ||
-            (set_pos <= 366 && set_pos >= 1)
+        unless (-366..366).include?(set_pos) && set_pos != 0
           raise ArgumentError, "Expecting number in [-366, -1] or [1, 366], got #{set_pos} (#{by_set_pos})"
         end
       end
 
       @by_set_pos = by_set_pos
-      replace_validations_for(:by_set_pos, by_set_pos && [Validation.new(by_set_pos, self)])
+      replace_validations_for(:by_set_pos, [Validation.new(by_set_pos, self)])
       self
     end
 
@@ -26,7 +19,6 @@ module IceCube
       attr_reader :rule, :by_set_pos
 
       def initialize(by_set_pos, rule)
-
         @by_set_pos = by_set_pos
         @rule = rule
       end
@@ -40,12 +32,11 @@ module IceCube
       end
 
       def validate(step_time, schedule)
-        start_of_month = TimeUtil.start_of_month step_time
-        end_of_month = TimeUtil.end_of_month step_time
+        start_of_month = step_time.beginning_of_month
+        end_of_month = step_time.end_of_month
 
-
-        new_schedule = IceCube::Schedule.new(TimeUtil.previous_month(step_time)) do |s|
-          s.add_recurrence_rule IceCube::Rule.from_hash(rule.to_hash.reject{|k, v| [:by_set_pos, :count, :until].include? k})
+        new_schedule = IceCube::Schedule.new(step_time - 1.month) do |s|
+          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :util)))
         end
 
         occurrences = new_schedule.occurrences_between(start_of_month, end_of_month)
@@ -63,7 +54,6 @@ module IceCube
           end
         end
       end
-
 
       def build_s(builder)
         builder.piece(:by_set_pos) << by_set_pos
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -36,7 +36,7 @@ module IceCube
         end_of_month = step_time.end_of_month
 
         new_schedule = IceCube::Schedule.new(step_time - 1.month) do |s|
-          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :util)))
+          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :until)))
         end
 
         occurrences = new_schedule.occurrences_between(start_of_month, end_of_month)
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -31,11 +31,12 @@ module IceCube
         true
       end
 
-      def validate(step_time, schedule)
+      def validate(step_time, start_time)
         start_of_month = step_time.beginning_of_month
         end_of_month = step_time.end_of_month
 
-        new_schedule = IceCube::Schedule.new(step_time - 1.month) do |s|
+        # Needs to start on the first day of the month
+        new_schedule = IceCube::Schedule.new(start_of_month.change(hour: step_time.hour, min: step_time.min, sec: step_time.sec)) do |s|
           s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :until)))
         end
 
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -3,7 +3,7 @@ module IceCube
   module Validations::MonthlyBySetPos
 
     def by_set_pos(*by_set_pos)
-      return by_set_pos([by_set_pos]) if by_set_pos.is_a?(Fixnum)
+      return by_set_pos([by_set_pos]) if by_set_pos.is_a?(Integer)
 
       unless by_set_pos.nil? || by_set_pos.is_a?(Array)
         raise ArgumentError, "Expecting Array or nil value for count, got #{by_set_pos.inspect}"
@@ -48,9 +48,7 @@ module IceCube
           s.add_recurrence_rule IceCube::Rule.from_hash(rule.to_hash.reject{|k, v| [:by_set_pos, :count, :until].include? k})
         end
 
-        puts step_time
         occurrences = new_schedule.occurrences_between(start_of_month, end_of_month)
-        p occurrences
         index = occurrences.index(step_time)
         if index == nil
           1
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -32,11 +32,12 @@ module IceCube
       end
 
       def validate(step_time, start_time)
-        start_of_month = step_time.beginning_of_month
-        end_of_month = step_time.end_of_month
+        start_of_month = TimeUtil.build_in_zone([step_time.year, step_time.month, 1, 0, 0, 0], step_time)
+        eom_date = Date.new(step_time.year, step_time.month, -1)
+        end_of_month = TimeUtil.build_in_zone([eom_date.year, eom_date.month, eom_date.day, 23, 59, 59], step_time)
 
         # Needs to start on the first day of the month
-        new_schedule = IceCube::Schedule.new(start_of_month.change(hour: step_time.hour, min: step_time.min, sec: step_time.sec)) do |s|
+        new_schedule = IceCube::Schedule.new(IceCube::TimeUtil.build_in_zone([start_of_month.year, start_of_month.month, start_of_month.day, step_time.hour, step_time.min, step_time.sec], start_of_month)) do |s|
           s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :until)))
         end
 
--- a/lib/ice_cube/validations/monthly_by_set_pos.rb
+++ b/lib/ice_cube/validations/monthly_by_set_pos.rb
@@ -36,24 +36,13 @@ module IceCube
       end
 
       def validate(step_time, start_time)
-        # Define the current month window so BYSETPOS is applied per month.
-        start_of_month = TimeUtil.build_in_zone([step_time.year, step_time.month, 1, 0, 0, 0], step_time)
-        eom_date = Date.new(step_time.year, step_time.month, -1)
-        end_of_month = TimeUtil.build_in_zone([eom_date.year, eom_date.month, eom_date.day, 23, 59, 59], step_time)
+        # Compute the interval bounds and build a filtered schedule that preserves
+        # implicit anchors while avoiding BYSETPOS/COUNT/UNTIL truncation.
+        start_of_month, end_of_month = Validations::BySetPosHelper.interval_bounds(:month, step_time)
+        new_schedule = Validations::BySetPosHelper.build_filtered_schedule(rule, start_time)
 
-        # Use the schedule start_time to preserve implicit date/time anchors.
-        new_schedule = IceCube::Schedule.new(start_time) do |s|
-          filtered_hash = rule.to_hash.reject { |key, _| [:by_set_pos, :count, :until].include?(key) }
-          # Avoid recursive BYSETPOS evaluation in the temporary schedule.
-          if filtered_hash[:validations]
-            filtered_hash[:validations] = filtered_hash[:validations].reject { |key, _| key == :by_set_pos }
-            filtered_hash.delete(:validations) if filtered_hash[:validations].empty?
-          end
-          s.add_recurrence_rule(IceCube::Rule.from_hash(filtered_hash))
-        end
-
-        # Build the full candidate set for this interval without COUNT/UNTIL,
-        # then map the selected occurrence to positive/negative positions.
+        # Build the full candidate set for this interval, then map the selected
+        # occurrence to positive/negative positions.
         occurrences = new_schedule.occurrences_between(start_of_month, end_of_month)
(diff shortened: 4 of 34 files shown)


davidstosik:
I can confirm this works as expected, this is fantastic! 👍

.gitignore
```diff
@@ -8,5 +8,8 @@
 /spec/reports/
 /tmp/
 
+# rubymine
+.idea
+
```
davidstosik: Is that necessary? It looks like an accidental addition that is not part of this PR.

lib/ice_cube/time_util.rb
```diff
@@ -198,6 +200,36 @@ def self.which_occurrence_in_month(time, wday)
       [nth_occurrence_of_weekday, this_weekday_in_month_count]
     end
 
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.start_of_month time
+      time.beginning_of_month
```
davidstosik: Why not reuse the same name?

lib/ice_cube/time_util.rb
```diff
@@ -198,6 +200,36 @@ def self.which_occurrence_in_month(time, wday)
...
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.end_of_month time
+      time.end_of_month
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.start_of_year time
+      time.beginning_of_year
```
davidstosik: Why not reuse the same name?

lib/ice_cube/time_util.rb
```diff
@@ -198,6 +200,36 @@ def self.which_occurrence_in_month(time, wday)
...
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.end_of_year time
+      time.end_of_year
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.previous_month time
+      time - 1.month
```
davidstosik: I think `time.last_month` exists?

lib/ice_cube/time_util.rb
```diff
@@ -198,6 +200,36 @@ def self.which_occurrence_in_month(time, wday)
...
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.previous_year time
+      time - 1.year
```
davidstosik: I think `time.last_year` exists?

lib/ice_cube/time_util.rb
```diff
@@ -198,6 +200,36 @@ def self.which_occurrence_in_month(time, wday)
...
+    def self.end_of_year time
+      time.end_of_year
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
+    def self.previous_month time
+      time - 1.month
+    end
+
+    # Use Activesupport CoreExt functions to manipulate time
```
davidstosik: I don't understand the point of adding those class methods to `TimeUtils`.
Wouldn't it be easier to just call `time.whatever` instead of `TimeUtils.whatever(time)` wherever we need those methods?

lib/ice_cube/validations/yearly_by_set_pos.rb
```diff
@@ -0,0 +1,89 @@
...
+
+          if @by_set_pos.include?(positive_set_pos) || @by_set_pos.include?(negative_set_pos)
+            0
+          else
+            1
+          end
+        end
+
+
+
```
davidstosik: [Nitpick] Too many empty lines?

spec/examples/by_set_pos_spec.rb
```diff
@@ -0,0 +1,29 @@
+require File.dirname(__FILE__) + '/../spec_helper'
+
+module IceCube
+
+  describe MonthlyRule, 'BYSETPOS' do
+    it 'should behave correctly' do
+      schedule = IceCube::Schedule.from_ical "RRULE:FREQ=MONTHLY;COUNT=4;BYDAY=WE;BYSETPOS=4"
+      schedule.start_time = Time.new(2015, 5, 28, 12, 0, 0)
+      expect(schedule.occurrences_between(Time.new(2015, 01, 01), Time.new(2017, 01, 01))).to eq([
+                                                                                                     Time.new(2015,6,24,12,0,0),
```
davidstosik: [Nitpick] Any way to avoid such long lines?

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
+module IceCube
+
+  module Validations::MonthlyBySetPos
+
+    def by_set_pos(*by_set_pos)
+      return by_set_pos([by_set_pos]) if by_set_pos.is_a?(Integer)
```
davidstosik: As you used the splat operator in the method's signature, I don't think it is possible for `by_set_pos` to be an `Integer`. This line thus seems unnecessary.

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
...
+
+      unless by_set_pos.nil? || by_set_pos.is_a?(Array)
+        raise ArgumentError, "Expecting Array or nil value for count, got #{by_set_pos.inspect}"
```
davidstosik: This validation seems unnecessary as well. `by_set_pos` **is always** an `Array`.

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
...
+      end
+      by_set_pos.flatten!
+      by_set_pos.each do |set_pos|
+        unless (set_pos >= -366 && set_pos <= -1) ||
```
davidstosik: How about `(-366..366).include(set_post) && set_pos != 0`?

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
...
+            (set_pos <= 366 && set_pos >= 1)
+          raise ArgumentError, "Expecting number in [-366, -1] or [1, 366], got #{set_pos} (#{by_set_pos})"
```
davidstosik: Should this error display only the invalid `set_pos`, instead of the complete `by_set_post` list?

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
...
+        end
+      end
+
+      @by_set_pos = by_set_pos
+      replace_validations_for(:by_set_pos, by_set_pos && [Validation.new(by_set_pos, self)])
```
davidstosik: If I understand correctly, `by_set_pos` cannot be `nil` or `false` here (it's an `Array`), in which case `by_set_pos && [Validation.new(by_set_pos, self)]` is equivalent to its right-side operand, `[Validation.new(by_set_pos, self)]`.

lib/ice_cube/validations/yearly_by_set_pos.rb
```diff
@@ -0,0 +1,89 @@
+module IceCube
+
+  module Validations::YearlyBySetPos
```
davidstosik: Most of my comments in `MonthlyBySetPos` might apply here as well.

davidstosik (commented):
I left comments to see if we can improve the code of this PR.
I'll try implementing them myself and link a diff here a bit later.

nehresma:
Hello @davidstosik, @NicolasMarlier provided this original PR.  I simply updated it to work with modern versions of Ruby.

dimerman:
@nehresma what's the status of this PR? I'd very much like to use the `setpos` functionality.

nehresma:
@dimerman It has been merged but I do not believe a new version of the gem has been released.  I've pointed my Gemfile to the latest git ref on master for the time being.  Hopefully someone will have a few minutes to cut a new gem release at some point so we wont have to point to specific refs.

nehresma:
@dimerman Err, no -- I'm an idiot.  I just realized it has NOT been merged.  I checked my Gemfile and I'm pointing to the ref with BYSETPOS support out on my fork -- not master on seejohnrun/ice_cube.

Sorry for the confusion.  I should have double checked before trusting my memory.

dimerman:
@nehresma no worries - I've been pointing my gemfile to your branch/repo. thanks for that!

However, I found the following case is not working right:
```
rule_a=FREQ=MONTHLY;COUNT=12;BYDAY=MO,TU,WE,TH,FR;BYSETPOS=-1`
rule_b=FREQ=MONTHLY;COUNT=12;BYDAY=MO,TU,WE,TH,FR;BYMONTHDAY=13,14,15;BYSETPOS=-1
```

when I create a `IceCube::Schedule` and add each rule individually, the occurrences are accurate. But if I create a schedule with both these rules, I get 18 occurrences instead of the expected 24.
I thought you might want to know. 🙌

mattbooks:
What's still holding this back — the PR feedback seems pretty minor? If that gets cleaned up can we get this merged?

mattbooks:
@seejohnrun It seems like several people need this and are using forks to get around it being missing. I'd be more than happy to help clean this up to avoid using a fork myself if it is otherwise merge-able.

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
...
+            0
+          else
+            1
+          end
+        end
+      end
+
+
+      def build_s(builder)
+        builder.piece(:by_set_pos) << by_set_pos
```
ovinix: Hello @nehresma 

Looks like this should be improved:
```
> IceCube::Rule.monthly(2).day(1,2,3,4,5).by_set_pos(2).to_s
"Every 2 months on Weekdays [2]"
```

pacso:
@nehresma / @mattbooks / @ovinix - If you can get this PR cleaned up and resolve the feedback and conflicts with `master`, I'll give it a proper review.

Naren1997:
Hello, looks like the gem already supporting what BYTSETPOS does but in different ical format. Example: If we want recurring for every 2 months 3rd Wednesday, ical format w.r.t the BYSETPOS is **FREQ=MONTHLY;INTERVAL=1;BYDAY=MO;BYSETPOST=3** which is same as **FREQ=MONTHLY;INTERVAL=1;BYDAY=3MO**, the latter one is already supported by the gem, so not sure why we still wants this PR.

davidstosik:
> not sure why we still wants this PR

I don’t use the gem anymore, but you might want to consider that some people use the gem to parse strings produced by another piece of software they don’t have control on, and this would require the whole specification to be supported. 😉

jhubert:
@avit Anything we can do to help get this merged?

dimerman:
@davidstosik what gem do you use instead?

davidstosik:
Nothing. 😬 
My need to handle schedules has vanished. 🙇🏻‍♂️

nehresma:
It's been over 4 years since I created this PR and over 7 years since Nicolas Marlier's initial commit with this support.  I have rebased it against current master and I've applied changes per the feedback received above.

If someone has time to review and merge, that would be most appreciated.  BYSETPOS is part of RFC 5545 and it would be great if this gem could finally get support merged in rather than just the "noop" that it currently is (https://github.com/ice-cube-ruby/ice_cube/blob/master/lib/ice_cube/parsers/ical_parser.rb#L77-L78).

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,75 @@
...
+      def dst_adjust?
+        true
+      end
+
+      def validate(step_time, schedule)
+        start_of_month = step_time.beginning_of_month
+        end_of_month = step_time.end_of_month
+
+        new_schedule = IceCube::Schedule.new(step_time - 1.month) do |s|
+          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :util)))
```
jkehres, line 39: Did you mean `:until` instead of `:util`?

nehresma: @jkehres Indeed I did!  Thanks for noticing this.

pacso:
I'll take a look. Am going to try to dedicate sometime this holiday season to get things tidied up and in better shape.

jkehres:
Trying this out for monthly frequencies with an interval > 1 seems to get stuck an infinite loop. I used the monthly test case from this PR and ran the following code, which defaults to an interval of 1, and everything works fine:

```ruby
schedule = IceCube::Schedule.from_ical "RRULE:FREQ=MONTHLY;COUNT=4;BYDAY=WE;BYSETPOS=4"
schedule.start_time = Time.new(2015, 5, 28, 12, 0, 0)
schedule.next_occurrence(Time.new(2015, 01, 01))
=> 2015-06-24 12:00:00 -0400
```

However, if I add an interval value > 1 then `next_occurrence` never returns:

```ruby
schedule = IceCube::Schedule.from_ical "RRULE:FREQ=MONTHLY;COUNT=4;BYDAY=WE;BYSETPOS=4;INTERVAL=2"
schedule.start_time = Time.new(2015, 5, 28, 12, 0, 0)
schedule.next_occurrence(Time.new(2015, 01, 01))
# method hangs
```

If I add a `puts` statement in `MonthlyBySetPos#Validation#validate` to log `step_time` and run the code above again I can see it just going infinitely into the future.

The same thing happens for yearly frequencies with an interval > 1.

lib/ice_cube/time_util.rb
```diff
@@ -1,5 +1,7 @@
-require "date"
-require "time"
+require 'date'
+require 'time'
+require 'active_support'
+require 'active_support/core_ext'
```
jkehres, line 4: This introduces a new dependency for the gem on `active_support`. The `gemspec` file probably needs to be updated accordingly. A non-Rails Ruby app won't necessarily have active support available so it needs to be listed as an explicit dependency.

jkehres:
I don't think this PR is sufficient to fully support BYSETPOS. It only deals with monthly and yearly frequencies. However,  playing around in the rrule.js demo [site](https://jakubroztocil.github.io/rrule/), BYSETPOS can have an effect on shorter frequencies too. For example these rule pairs each return different results with and without BYSETPOS:

```
RRULE:FREQ=WEEKLY;COUNT=15;INTERVAL=2;WKST=MO;BYDAY=MO,TU
RRULE:FREQ=WEEKLY;COUNT=15;INTERVAL=2;WKST=MO;BYDAY=MO,TU;BYSETPOS=1
```

```
RRULE:FREQ=DAILY;COUNT=15;INTERVAL=2;WKST=MO;BYHOUR=1,2
RRULE:FREQ=DAILY;COUNT=15;INTERVAL=2;WKST=MO;BYHOUR=1,2;BYSETPOS=1
```

```
RRULE:FREQ=HOURLY;COUNT=15;INTERVAL=2;WKST=MO;BYMINUTE=1,2
RRULE:FREQ=HOURLY;COUNT=15;INTERVAL=2;WKST=MO;BYMINUTE=1,2;BYSETPOS=1
```

```
RRULE:FREQ=MINUTELY;COUNT=15;INTERVAL=2;WKST=MO;BYSECOND=1,2
RRULE:FREQ=MINUTELY;COUNT=15;INTERVAL=2;WKST=MO;BYSECOND=1,2;BYSETPOS=1
```

nehresma:
Hey @jkehres I see what you mean regarding the infinite loop when combining interval and bysetpos.  I'm looking into it.

Also, you're right regarding the use of bysetpos and other intervals.  I'll look into adding support for those as well.

@NicolasMarlier is the original author of this work.  I've been shepherding it through bitrot issues over the years and applying feedback.

Thanks for your time evaluating this.

lib/ice_cube/time_util.rb
nehresma: @jkehres Thanks, I've fixed this now.  These were a holdover from the original commit in 2016 but TimeUtil was refactored since then to no longer need ActiveSupport for time arithmetic.

nehresma:
@jkehres I've fixed the interval use for month and year frequencies with bysetpos.  I ran out of time for adding support for the other frequencies yet.  I'll add support for those after Christmas.

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,76 @@
...
+      def type
+        :day
+      end
+
+      def dst_adjust?
+        true
+      end
+
+      def validate(step_time, start_time)
+        start_of_month = step_time.beginning_of_month
```
jkehres, line 35: The `beginning_of_month` method is not available in vanilla Ruby. It's from active support (defined [here](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/date_and_time/calculations.rb#L125)). I tried testing this PR out in a non-Rails IRB session and I get an undefined method error here because active support is not available. This PR either needs to add a dependency on the active support gem or it needs re-implement `beginning_of_month` with only vanilla Ruby. I'm not a maintainer of this gem (just an interested party that would like to help get this functionality merged) but I suspect that adding a new dependency on active support may not be something they'd be willing to accept since this gem currently has no dependencies.

lib/ice_cube/validations/yearly_by_set_pos.rb
```diff
@@ -0,0 +1,77 @@
...
+      def type
+        :day
+      end
+
+      def dst_adjust?
+        true
+      end
+
+      def validate(step_time, start_time)
+        start_of_year = step_time.beginning_of_year
```
jkehres, line 37: Ditto my comment from the previous file regarding `beginning_of_month`. The `beginning_of_year` method is only available via active support (defined [here](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/date_and_time/calculations.rb#L179)).

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,76 @@
...
+        end_of_month = step_time.end_of_month
+
+        # Needs to start on the first day of the month
+        new_schedule = IceCube::Schedule.new(start_of_month.change(hour: step_time.hour, min: step_time.min, sec: step_time.sec)) do |s|
```
jkehres, line 39: The `change` method is also from active support (defined [here](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/time/calculations.rb#L138)).

lib/ice_cube/validations/monthly_by_set_pos.rb
jkehres: Without using active support, I think this could be:

```ruby
start_of_month = IceCube::TimeUtil.build_in_zone([step_time.year, step_time.mo
nth, 1, 0, 0, 0], step_time)
```

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,76 @@
...
+        :day
+      end
+
+      def dst_adjust?
+        true
+      end
+
+      def validate(step_time, start_time)
+        start_of_month = step_time.beginning_of_month
+        end_of_month = step_time.end_of_month
```
jkehres, line 36: Without active support, this might work:

```ruby
last_day = IceCube::TimeUtil.days_in_month(step_time)
end_of_month = IceCube::TimeUtil.build_in_zone([step_time.year, step_time.mont
h, last_day, 23, 59, 59], step_time)
```

lib/ice_cube/validations/monthly_by_set_pos.rb
jkehres: Without active support:

```ruby
IceCube::TimeUtil.build_in_zone([start_of_month.year, start_of_month.month, start_of_month.day, step_time.hour, step_time.min, step_time.sec], start_of_month)
```

lib/ice_cube/validations/yearly_by_set_pos.rb
jkehres: Without active support:

```ruby
start_of_month = IceCube::TimeUtil.build_in_zone([step_time.year, 1, 1, 0, 0, 0], step_time)
```

lib/ice_cube/validations/yearly_by_set_pos.rb
```diff
@@ -0,0 +1,77 @@
...
+        end_of_year = step_time.end_of_year
```
jkehres, line 38: Without active support:

```ruby
end_of_year = IceCube::TimeUtil.build_in_zone([step_time.year, 12, 31, 23, 59, 59], step_time)
```

lib/ice_cube/validations/monthly_by_set_pos.rb
jkehres: Inspiration for this approach was taken from the existing `beginning_of_date` method (defined [here](https://github.com/ice-cube-ruby/ice_cube/blob/master/lib/ice_cube/time_util.rb#L137)).

jkehres: Maybe it's even worth creating `beginning_of_xxx` and `end_of_xxx` helpers in `IceCube::TimeUtil` that are equivalent to their active support counterparts and then used them in the code here.

nehresma: Nice catch.  Pushed 621bf3db42b20b38563da1cb70ca202bc6ce6f4a to resolve this.

nehresma:
I've pushed bysetpos support for weekly frequencies too.  I don't think bysetpos makes sense with frequencies of daily, hourly, minutely, or secondly.  RFC 5545 makes no mention of them, and indeed all of it's examples are with monthly and weekly frequencies.  If someone can think of an example where using bysetpos with one of these other frequencies makes sense, let me know.

spec/examples/to_yaml_spec.rb
```diff
@@ -78,7 +78,7 @@ module IceCube
     end
 
     it "should be able to make a round-trip to YAML with .day_of_year" do
-      schedule1 = Schedule.new(Time.now)
+      schedule1 = Schedule.new(Time.zone.now)
```
jkehres, line 81: `Time.zone` is another active support [extension](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/time/zones.rb#L14). The stock `Time` class in Ruby has no `zone` method.

jkehres:
> I've pushed bysetpos support for weekly frequencies too. I don't think bysetpos makes sense with frequencies of daily, hourly, minutely, or secondly. RFC 5545 makes no mention of them, and indeed all of it's examples are with monthly and weekly frequencies. If someone can think of an example where using bysetpos with one of these other frequencies makes sense, let me know.

I believe BYSETPOS support is still needed for daily, hourly, and minutely frequencies to be compliant with the RFC, which I think should be the goal of this PR. As I mentioned in my comment [here](https://github.com/ice-cube-ruby/ice_cube/pull/449#issuecomment-1363384247), you can see BYSETPOS having an effect for these other frequencies using rrule.js - another library that implements RFC 5545 (itself a port of the python-dateutil library). If ice_cube and these other libraries all properly implement the spec, they should all give the same answer for the same RRULE. We could use that fact to aid in testing this PR to ensure we have a compliant implementation of the RFC. We could us one of these other libraries to generate a more exhaustive set of test inputs and outputs for RRULEs that use BYSETPOS and, thus, increase the number of tests for this new code along with our confidence in it.

Secondly, even though the RFC gives no examples for these other frequencies with BYSETPOS, I think you can derive that they are necessary from the details in the spec.

From the description of BYSETPOS (emphasis mine):

> The BYSETPOS rule part specifies a COMMA-separated list of values that corresponds to **the nth occurrence within the set of recurrence instances** specified by the rule.  BYSETPOS operates on **a set of recurrence instances in one interval of the recurrence rule**.

My understanding of the spec is that "interval" in this context means one unit of the frequency regardless of the INTERVAL rule part so one year for yearly, one month for monthly, etc. If there are multiple recurrences instances within one interval (e.g. year, month, etc.), then BYSETPOS has an effect and selects the nth one. Normally, there is only a single recurrence instance within an interval. The way you get more than one recurrence instance in an interval is by adding another BYxxx rule part.

Again, from the spec (emphasis mine):

> BYxxx rule parts modify the recurrence in some manner.  BYxxx rule parts for a period of time that is the same or greater than the frequency generally reduce or limit the number of occurrences of the recurrence generated.  For example, "FREQ=DAILY;BYMONTH=1" reduces the number of recurrence instances from all days (if BYMONTH rule part is not present) to all days in January. **BYxxx rule parts for a period of time less than the frequency generally increase or expand the number of occurrences of the recurrence.** For example, "FREQ=YEARLY;BYMONTH=1,2" increases the number of days within the yearly recurrence set from 1 (if BYMONTH rule part is not present) to 2.

Furthermore, this table from the spec shows how each BYxxx rule part interacts with a given frequency and whether it expands or limits the number of recurrences instances (emphasis mine):

```
   +----------+--------+--------+-------+-------+------+-------+------+
   |          |SECONDLY|MINUTELY|HOURLY |DAILY  |WEEKLY|MONTHLY|YEARLY|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYMONTH   |Limit   |Limit   |Limit  |Limit  |Limit |Limit  |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYWEEKNO  |N/A     |N/A     |N/A    |N/A    |N/A   |N/A    |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYYEARDAY |Limit   |Limit   |Limit  |N/A    |N/A   |N/A    |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYMONTHDAY|Limit   |Limit   |Limit  |Limit  |N/A   |Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYDAY     |Limit   |Limit   |Limit  |Limit  |Expand|Note 1 |Note 2|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYHOUR    |Limit   |Limit   |Limit  |Expand |Expand|Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYMINUTE  |Limit   |Limit   |Expand |Expand |Expand|Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYSECOND  |Limit   |Expand  |Expand |Expand |Expand|Expand |Expand|
   +----------+--------+--------+-------+-------+------+-------+------+
   |BYSETPOS  |Limit   |Limit   |Limit  |Limit  |Limit |Limit  |Limit |
   +----------+--------+--------+-------+-------+------+-------+------+
```

For secondly, you can see that no BYxxx rule part will expand the set of recurrence instances so BYSETPOS won't have an effect. However, for minutely, hourly, and daily frequencies you can see that some BYxxx rule parts will expand the set of recurrences, which would allow BYSETPOS to select the nth one in an interval and thus effect the final set of recurrence instances.

spec/examples/by_set_pos_spec.rb
```diff
@@ -0,0 +1,103 @@
+require File.dirname(__FILE__) + '/../spec_helper'
+
+module IceCube
+  describe WeeklyRule, 'BYSETPOS' do
+    it 'should behave correctly' do
+      # Weekly on Monday, Wednesday, and Friday with the week starting on Wednesday, the last day of the set
+      schedule = IceCube::Schedule.from_ical("RRULE:FREQ=WEEKLY;COUNT=4;WKST=WE;BYDAY=MO,WE,FR;BYSETPOS=-1")
```
jkehres, line 7: For each of the frequencies, it would be nice to have the following test cases:

- With BYxxx that expands the set of recurrence instances...
  - BYSETPOS with a positive value
  - BYSETPOS with a negative value
  - BYSETPOS with multiple positive values
  - BYSETPOS with multiple negative values
  - BYSETPOS with multiple values both positive and negative
  - BYSETPOS with multiple repeated values - spec does not define behaviour; rrule.js does some weird things and outputs the same occurrence multiple times (bug or maybe just undefined behaviour?)
  - BYSETPOS with a value that matches in some intervals but not others (e.g. selecting the 5th Monday of every month since some months will have only 4 Mondays)
  - BYSETPOS with a positive value beyond than the number of recurrence instances in any interval (e.g. the 10th Monday of every month); rrule.js returns an empty set of occurrences
  - BYSETPOS with a negative value bigger than the number of recurrence instances in any set (e.g. the -10th Monday of every month); rrule.js returns a non-empty set of occurrences, which is weird (bug or maybe just undefined behaviour?)
- BYSETPOS with BYxxx that limits the set of recurrence instances instead of expanding it
- BYSETPOS with no BYxxx - spec says BYSETPOS "MUST only be used in conjunction with another BYxxx rule part" so not sure if this should be an exception or an empty set of occurrences returned; rrule.js does the latter
- BYSETPOS with multiple BYxxx rules parts that expand the recurrence instances
- BYSETPOS with multiple BYxxx rules parts that both expand and limit the recurrence instances
- BYSETPOS with a value outside the allowed range (covered in the `from_ical` tests but can you also construct a schedule using BYSETPOS without `from_ical`? if not, should that be a feature of this PR?)

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -75,7 +75,7 @@ def self.rule_from_ical(ical)
         when "BYYEARDAY"
           validations[:day_of_year] = value.split(",").map(&:to_i)
         when "BYSETPOS"
-          # noop
+          params[:validations][:by_set_pos] = value.split(',').collect(&:to_i)
```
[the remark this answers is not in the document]
jkehres: Also, existing code uses double quotes instead of single quotes for string in `split` and `collect` instead of `map`. Picky but just for consistency.

lib/ice_cube/time_util.rb
```diff
@@ -1,5 +1,5 @@
-require "date"
-require "time"
+require 'date'
+require 'time'
```
jkehres, line 2: Changing the type of quotes seems unnecessary. Double quotes are consistent with the existing style.

lib/ice_cube/validations/weekly_by_set_pos.rb
```diff
@@ -0,0 +1,91 @@
...
+
+        # Needs to start on the first day of the week at the step_time's hour, min, sec
+        start_of_week_adjusted = IceCube::TimeUtil.build_in_zone(
+          [
+            start_of_week_date.year, start_of_week_date.month, start_of_week_date.day,
+            step_time.hour, step_time.min, step_time.sec
+          ], step_time
+        )
+        new_schedule = IceCube::Schedule.new(start_of_week_adjusted) do |s|
+          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :until)))
```
jkehres, line 57: `except` is another active support method:

https://github.com/rails/rails/blob/v7.0.4/activesupport/lib/active_support/core_ext/hash/except.rb#L12

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,77 @@
...
+      end
+
+      def validate(step_time, start_time)
+        start_of_month = TimeUtil.build_in_zone([step_time.year, step_time.month, 1, 0, 0, 0], step_time)
+        eom_date = Date.new(step_time.year, step_time.month, -1)
+        end_of_month = TimeUtil.build_in_zone([eom_date.year, eom_date.month, eom_date.day, 23, 59, 59], step_time)
+
+        # Needs to start on the first day of the month
+        new_schedule = IceCube::Schedule.new(IceCube::TimeUtil.build_in_zone([start_of_month.year, start_of_month.month, start_of_month.day, step_time.hour, step_time.min, step_time.sec], start_of_month)) do |s|
+          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :until)))
```
jkehres, line 41: Ditto my [comment](https://github.com/ice-cube-ruby/ice_cube/pull/449/files#r1067495750) about `except`.

lib/ice_cube/validations/yearly_by_set_pos.rb
```diff
@@ -0,0 +1,77 @@
...
+        true
+      end
+
+      def validate(step_time, start_time)
+        start_of_year = TimeUtil.build_in_zone([step_time.year, 1, 1, 0, 0, 0], step_time)
+        end_of_year = TimeUtil.build_in_zone([step_time.year, 12, 31, 23, 59, 59], step_time)
+
+        # Needs to start on the first day of the year
+        new_schedule = IceCube::Schedule.new(IceCube::TimeUtil.build_in_zone([start_of_year.year, start_of_year.month, start_of_year.day, step_time.hour, step_time.min, step_time.sec], start_of_year)) do |s|
+          s.add_recurrence_rule(IceCube::Rule.from_hash(rule.to_hash.except(:by_set_pos, :count, :until)))
```
jkehres, line 42: Ditto my [comment](https://github.com/ice-cube-ruby/ice_cube/pull/449/files#r1067495750) about `except`.

jkehres:
Using BYSECOND with MONTHLY does not work. I tried this:

```ruby
ical = "RRULE:FREQ=MONTHLY;COUNT=4;BYSECOND=1,2,3,4;BYSETPOS=2"
schedule = IceCube::Schedule.from_ical(ical)
schedule.start_time = Time.new(2015, 5, 28, 12, 0, 0)
schedule.occurrences_between(Time.new(2015, 01, 01), Time.new(2017, 01, 01))
```

and got an empty array. I expected to get:

```
2015-05-28 12:00:02
2015-06-28 12:00:02
2015-07-28 12:00:02
2015-08-28 12:00:02
```

I suspect there is an assumption in the code that you will only use BYDAY with MONTHLY and not other BYxxx rule parts.

rgalanakis:
First I want to thank @nehresma @jkehres and others for all the work on this.

> I believe BYSETPOS support is still needed for daily, hourly, and minutely frequencies to be compliant with the RFC, which I think should be the goal of this PR.

The current code already doesn't implement the RFC, and it fails to do so silently (just spent over an hour finding the this as the source of a bug). I think partial RFC compliance is better than none. And the users who use daily/hourly/minutely frequencies would continue to see the current unimplemented behavior. I know I wouldn't be alone in appreciating getting this work merged.

pacso:
Thanks for the review @jkehres. Can we get the tests running/passing?

jorgemanrubia:
Thank you so much for your work here. I'd love if we can get this merged. Apple Calendar feeds use `BYSETPOS` when selecting a day of the month. And there is no direct equivalent for the negative values (starting from the end). So this would be super handy.

pacso:
@jorgemanrubia - Since nothing has moved on this PR for a long time ... it might be easiest if you can fork it, get the tests passing and then we can move forwards? I have struggled for time recently otherwise would have done this myself.

Skulli:
Monthly/Yearly would be more than enough for me.

lib/ice_cube/validations/monthly_by_set_pos.rb
```diff
@@ -0,0 +1,85 @@
...
+        true
+      end
+
+      def validate(step_time, schedule)
+        start_of_month = TimeUtil.start_of_month step_time
+        end_of_month = TimeUtil.end_of_month step_time
+
+
+        new_schedule = IceCube::Schedule.new(TimeUtil.previous_month(step_time)) do |s|
+          s.add_recurrence_rule IceCube::Rule.from_hash(rule.to_hash.reject{|k, v| [:by_set_pos, :count, :until].include? k})
```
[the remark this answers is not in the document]
nehresma: Yes but ActiveSupport isn't a runtime requirement of this gem.

nehresma:
Hmm, yes you are right -- I would expect the 11th to be the first event.  I haven't tested this particular event with this branch of ice_cube yet.  I'll take a closer look at this, fix it, and add test coverage for this to verify it.  Thanks!!!

rgalanakis:
Thanks, confirmed working on my end. Appreciate the quick response as it saved some git headaches (I stupidly took a dependency directly on your fork)

nehresma:
@pacso I can definitely address the linting issues in the code that was changed in this pull request, but I'll leave the code that wasn't touched here alone.  It'll make merging easier.

pacso:
Thanks everybody for all of your hard work! I'll see what else is ready to merge this week and then get a new release out.

AntoineGirard:
Thanks all for your work on this feature 🙇🏻

It’s a pleasure to discover that we need to support BYSETOPS this week, and then you merge this long work last week.

Long living open source!

nehresma:
@pacso any chance you could cut a release to rubygems.org?

pacso:
Yes, I will do. I just wanted to see whether there were any other PRs that ought to also be included before I did it.

timwis:
Hi @pacso we've just run into this as well. We even set off trying to develop a patch for it, only to realise it was already fixed in main. Any chance you can publish a new release on rubygems so we don't have to pin to main? Thanks so much! 🙏🏻

PR #454, opened 2018-08-25
Update README with RDATE support

thefliik:
I believe 0.16.3 added RDATE support ~~making ical serialization fully supported (?)~~

Update:
Looks like parsing time zone information still isn't supported at the moment. Added note about that.

Update 2:
Also added time zone parsing support in #455.


seejohnrun:
Thanks!

PR #455, opened 2018-08-25
Add timezone parsing / serializing

thefliik:
This PR updates the `full_tz` branch with correct timezone parsing / serializing. I'll add some comments in the code to explain some of the changes. It uses the `responds_to?(:time_zone)` method you already had in place to check for `ActiveSupport`.

### DIFF
--- a/lib/ice_cube/builders/ical_builder.rb
+++ b/lib/ice_cube/builders/ical_builder.rb
@@ -37,11 +37,13 @@ module IceCube
 
     def self.ical_format(time, force_utc)
       time = time.dup.utc if force_utc
+
       if time.utc?
         ":#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%SZ')}" # utc time
+      elsif time.respond_to?(:time_zone)
+        ";TZID=#{time.time_zone.name}:#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%S')}" # time zone specified
       else
-        time_zone = time.respond_to?(:time_zone) ? time.time_zone.name : time.zone
-        ";TZID=#{time_zone}:#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%S')}" # local time specified
+        ":#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%S')}" # local time specified
       end
     end
 
--- a/lib/ice_cube/parsers/ical_parser.rb
+++ b/lib/ice_cube/parsers/ical_parser.rb
@@ -5,17 +5,19 @@ module IceCube
       ical_string.each_line do |line|
         (property, value) = line.split(':')
         (property, tzid) = property.split(';')
+        (_, time_zone) = tzid.split('=') if tzid
+
         case property
         when 'DTSTART'
-          data[:start_time] = TimeUtil.deserialize_time(value)
+          data[:start_time] = TimeUtil.deserialize_time({time: value, zone: time_zone})
         when 'DTEND'
-          data[:end_time] = TimeUtil.deserialize_time(value)
+          data[:end_time] = TimeUtil.deserialize_time({time: value, zone: time_zone})
         when 'RDATE'
           data[:rtimes] ||= []
-          data[:rtimes] += value.split(',').map { |v| TimeUtil.deserialize_time(v) }
+          data[:rtimes] += value.split(',').map { |v| TimeUtil.deserialize_time({time: v, zone: time_zone}) }
         when 'EXDATE'
           data[:extimes] ||= []
-          data[:extimes] += value.split(',').map { |v| TimeUtil.deserialize_time(v) }
+          data[:extimes] += value.split(',').map { |v| TimeUtil.deserialize_time({time: v, zone: time_zone}) }
         when 'DURATION'
           data[:duration] # FIXME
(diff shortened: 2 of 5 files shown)


lib/ice_cube/builders/ical_builder.rb
```diff
@@ -37,11 +37,13 @@ def self.ical_utc_format(time)
...
       time = time.dup.utc if force_utc
+
       if time.utc?
         ":#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%SZ')}" # utc time
+      elsif time.respond_to?(:time_zone)
+        ";TZID=#{time.time_zone.name}:#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%S')}" # time zone specified
       else
-        time_zone = time.respond_to?(:time_zone) ? time.time_zone.name : time.zone
-        ";TZID=#{time_zone}:#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%S')}" # local time specified
+        ":#{IceCube::I18n.l(time, format: '%Y%m%dT%H%M%S')}" # local time specified
```
thefliik: As shown in the spec, dates in local / floating time should be serialized in the form `YYYYMMDDTHHMMSS`. Previously they were serialized with `TZID=${time.zone}`.

While I don't think `TZID=${time.zone}` is invalid (if only because, as the spec states, there is no specified format for timezones), it isn't a *local* time and it also doesn't produce time zones which ActiveSupport expects.

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -5,17 +5,19 @@ def self.schedule_from_ical(ical_string, options = {})
...
-          data[:end_time] = TimeUtil.deserialize_time(value)
+          data[:end_time] = TimeUtil.deserialize_time({time: value, zone: time_zone})
         when 'RDATE'
           data[:rtimes] ||= []
-          data[:rtimes] += value.split(',').map { |v| TimeUtil.deserialize_time(v) }
+          data[:rtimes] += value.split(',').map { |v| TimeUtil.deserialize_time({time: v, zone: time_zone}) }
         when 'EXDATE'
           data[:extimes] ||= []
-          data[:extimes] += value.split(',').map { |v| TimeUtil.deserialize_time(v) }
+          data[:extimes] += value.split(',').map { |v| TimeUtil.deserialize_time({time: v, zone: time_zone}) }
```
thefliik: I updated the `TimeUtil.deserialize_time` to handle a hash argument where the `:zone` key might equal `nil`.

Also, while the `TimeUtil.deserialize_time` method could already handle parsing times with zone information, the ical_parser never sent zone information to the `deserialize_time()` method.

lib/ice_cube/time_util.rb
```diff
@@ -109,10 +109,10 @@ def self.deserialize_time(time_or_hash)
       when Time, Date
         time_or_hash
       when DateTime
-        Time.local(time.year, time.month, time.day, time.hour, time.min, time.sec)
+        Time.local(time_or_hash.year, time_or_hash.month, time_or_hash.day, time_or_hash.hour, time_or_hash.min, time_or_hash.sec)
```
thefliik: So I *think* that this was a bug. Not sure when, if ever, the `when DateTime` condition is triggered, but before you had `Time.local(time.year ... ` but the variable `time` doesn't appear to be defined anywhere.

I *think* this was a copy-paste error from the `serialize_time()` method above.

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -5,17 +5,19 @@ def self.schedule_from_ical(ical_string, options = {})
       ical_string.each_line do |line|
         (property, value) = line.split(':')
         (property, tzid) = property.split(';')
+        (_, time_zone) = tzid.split('=') if tzid
```
iainbeeston: I believe this won't always work - looking at the spec, a component may have many parameters, of the form `<KEY>=<VALUE>`, separated by semicolons. If the key is `TZID`, then the value is the time zone that you want. But it there are other keys too.

For example, this is a valid ical line: `DTSTART;VALUE=DATE:20200525`. The code here would parse this and get the time zone `DATE`, which is not correct. The `VALUE` parameter tells us that the value (after the colon - `20200525` should be parsed as a date.

Another example would be `DTSTART;VALUE=DATE-TIME;TZID:US-Eastern:19980119T020000`. Here we do have a time zone, but it would be discarded by the code we have, and instead the time zone would be `DATE-TIME`.

thefliik: @iainbeeston you're correct.

This might fix it:

```ruby
(property, *params) = property.split(';')
params.map! { |p| p.split('=') }
time_zone = params.select { |p| p&.first&.upcase == "TZID" }.first&.last
value_type = params.select { |p| p&.first&.upcase == "VALUE" }.first&.last
```

`TimeUtil.deserialize_time` should also be updated so that the `Hash` case accepts a `type` key and handles both `"DATE"` and `"DATE-TIME"` cases (I don't think the `"TIME"` value is applicable to this library).

suzumejakku:
Hi,
It would be very nice to have this official support of timezone parsing; can the author init a merge ?

PR #458, opened 2018-09-28
Add Dutch (NL) translations

kogre:
Thanks for your great library.

Here are translations for Dutch, hope these can get merged!


seejohnrun:
Thank you!

PR #459, opened 2018-10-02
Drop support for old ruby and ruby rails versions

gssbzn:
Versions of Ruby before 2.3 have reach EOL and should not be supported anymore
Versions of Rails before 4.2 have reach EOL and should not be supported anymore (4.2 soon to follow)

While dropping support to old versions bring the code base to more current standard styles and practices
- Drop deprecated `has_rdoc` and `test_files` from gemspec
- Freeze constants to avoid manipulation of arrays or hashes
- Update hash style to drop rockets

And others

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,32 +10,12 @@ branches:
   only:
     - master
 rvm:
-  - 1.9.3
-  - 2.0
-  - 2.1.10
-  - 2.2.6
-  - 2.3.3
-  - 2.4.0
+  - 2.3.7
+  - 2.4.4
+  - 2.5.1
   - ruby-head
 env:
-  - RAILS_VERSION='~> 3.2'
   - RAILS_VERSION='~> 4.2'
   - RAILS_VERSION='~> 5.0'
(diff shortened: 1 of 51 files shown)


iainbeeston:
Another advantage of this is that bundler 2.0 is out now, which only works on ruby 2.3 and greater. The build on master is failing right now for older ruby versions, because travis is installing bundler 2.0 and then erroring out because the ruby version is incompatible.

jopotts:
@avit Thanks for all the amazing work on ice_cude over the years. I think this PR would be a good one to kickstart some updates? I'm looking for solid ruby 2.5 support so wondering the best way forwards on that?

jopotts:
It's a shame to see this PR closed. This gem needs some core updates to keep it working on newer rubies.

@avit & @seejohnrun is there some way you could open this up to more community maintainers?

gssbzn:
I'm no longer actively working with this gem or Ruby in general and considering the PR was open for  I decided to close it and let someone else give it a try

Issue #462, opened 2018-12-12
i18n v1.2.0 has a breaking change


lucascaton:
Hi there.

**i18n v1.2.0** was [released one day ago](https://github.com/svenfuchs/i18n/releases/tag/v1.2.0) and I've noticed one change that affects `ice_cube`: 

Locale's **integer** keys are being converted to **symbols**.

#### Before (`i18n v1.1.1`):

```ruby
require 'bundler/inline'

gemfile do
  gem 'i18n', '1.1.1'
  gem 'ice_cube'
end

puts IceCube::I18n.t('ice_cube.integer.ordinals')
# => {:default=>"th", 1=>"st", 2=>"nd", 3=>"rd", 11=>"th", 12=>"th", 13=>"th"}
```

#### After (`i18n v1.2.0`):

```ruby
require 'bundler/inline'

gemfile do
  gem 'i18n', '1.2.0'
  gem 'ice_cube'
end

puts IceCube::I18n.t('ice_cube.integer.ordinals')
# => {:default=>"th", :"1"=>"st", :"2"=>"nd", :"3"=>"rd", :"11"=>"th", :"12"=>"th", :"13"=>"th"}
```

---

I'm not sure if this is a [i18n](https://github.com/svenfuchs/i18n/) bug or an expected change, but in any case several methods from `lib/ice_cube/builders/string_builder.rb` has stopped working. This is one example:

```ruby
puts IceCube::StringBuilder.nice_number(1)
# [i18n v.1.1.1] => returns `1st`
# [i18n v.1.2.0] => returns `1th`
```

CodingAnarchy:
Well, this explains why I was seeing `Monthly on the 2th Thursday and -1th to last Thursday` all of a sudden.

CodingAnarchy:
It looks like this is the breaking line:
```ruby
 f = IceCube::I18n.t('ice_cube.string.format')[type] ? type : 'default'
```
I believe I can fix this based on the info in svenfuchs/i18n#422.

radar:
Apologies for the issue. Please be aware that this change will be reverted in i18n 1.2.1, which will be released within the next day.

radar:
Could you please test this branch with and without your patch: https://github.com/svenfuchs/i18n/pull/444

I believe it will support both behaviours.

CodingAnarchy:
[...]

PR #463, opened 2018-12-14
Fix for I18n 1.2.0 changing handling of integer keys 

CodingAnarchy:
This fix restores the handling of ordinal internationalization by adapting to the changes made for `i18n` in svenfuchs/i18n#422.

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -1,6 +1,7 @@
 require "bundler/setup"
 require 'ice_cube'
 require 'timeout'
+require 'i18n'
 
 begin
   require 'simplecov'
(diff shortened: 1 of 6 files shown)


CodingAnarchy:
Fixes #462

spec/spec_helper.rb
```diff
@@ -9,6 +10,8 @@
   # okay
 end
 
+I18n.config.available_locales = [:en, :es, :de, :fr, :ja, :nl, :"pt-BR", :ru, :sv]
```
saiqulhaq: why there are so many available locales here?
but in [here](https://github.com/seejohnrun/ice_cube/pull/463/files#diff-05234c22ea23d8af294b12812e505764R5) you write only :en

PR #465, opened 2019-01-12
Added support for parsing VEVENT from iCal format

iainbeeston:
BACKGROUND:
-----

I have an ical file I'd like to parse with ice cube. It has no recurring events, just a series of one off events. Those events are stored in the file as `VEVENT`s, where each event is a block of lines all describing a single event, starting with `BEGIN:VEVENT` and ending with `END:VEVENT` (more detail [here](https://www.kanzaki.com/docs/ical/vevent.html)).

CURRENT BEHAVIOUR
-------

Icecube can't parse these events. It sees the `DTSTART` and `DTEND` inside the `VEVENT`, but applies them to the schedule.

EXPECTED BEHAVIOUR
-----

I would have expected it to add each `VEVENT` as it's own recurrence time in the new schedule.

SOLUTION
-----

I've refactored the ical parser such that it acts as a state machine. By default it works as it did before. However, if it sees a `BEGIN:VEVENT` line it goes into an event parsing mode, where it parses subsequent lines as events. Once it sees an `END:VEVENT` line, it goes back to the regular mode. When parsing events it adds a new recurrence time every time it sees a `DTSTART`.

NOTES
----

* To keep the code simple, if there are multiple `DTSTART` lines within an event, it will add a new recurrence time for each of them. A `VEVENT` with more than one `DTSTART` is invalid, so this shouldn't happen in practice.
* `VEVENT`s can have an end date (ie. `DTEND`) as well as a start date (`DTSTART`). I couldn't see a way of adding a duration to a recurrence time, so I've ignored the end dates for now.


iainbeeston:
The build is failing because travis is trying to use bundler 2.0 with ruby < 2.3. #459 should fix that

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -2,28 +2,57 @@ module IceCube
   class IcalParser
     def self.schedule_from_ical(ical_string, options = {})
       data = {}
+      parser = :parse_line
       ical_string.each_line do |line|
         (property, value) = line.split(":")
-        (property, _tzid) = property.split(";")
-        case property
+        (property, tzid) = property.split(";")
```
pacso, line 8:
```suggestion
        (property, _tzid) = property.split(";")
```

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -2,28 +2,57 @@ module IceCube
...
         when "DURATION"
-          data[:duration] # FIXME
+          nil # FIXME
         when "RRULE"
-          data[:rrules] ||= []
-          data[:rrules] += [rule_from_ical(value)]
+          [:parse_line, :rrules, [rule_from_ical(value)]]
+        when "BEGIN"
+          [:parse_vevent_line] if value.chomp == "VEVENT"
         end
```
pacso, line 42:
```suggestion
      end
```

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -2,28 +2,57 @@ module IceCube
...
-      end
-      Schedule.from_hash data
+
+      result || [:parse_line]
+    end
+
+    def self.parse_vevent_line(property, value)
+      result = case property
+        when 'DTSTART'
```
pacso, line 49:
```suggestion
        when "DTSTART"
```

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -2,28 +2,57 @@ module IceCube
...
+          [:parse_vevent_line, :rtimes, [TimeUtil.deserialize_time(value)]]
+        when 'END'
+          [:parse_line] if value.chomp == 'VEVENT'
```
pacso, line 52:
```suggestion
          [:parse_line] if value.chomp == "VEVENT"
```

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -2,28 +2,57 @@ module IceCube
...
+        end
```
pacso, line 53:
```suggestion
      end
```

spec/examples/from_ical_spec.rb
```diff
@@ -378,13 +389,44 @@ def sorted_ical(ical)
       end
     end
 
+    describe 'one-off dates' do
+      it 'handles a specific day as a recurrence time' do
```
pacso, line 393:
```suggestion
      it "handles a specific day as a recurrence time" do
```

pacso (approved):

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -2,28 +2,57 @@ module IceCube
...
       data = {}
+      parser = :parse_line
       ical_string.each_line do |line|
         (property, value) = line.split(":")
         (property, _tzid) = property.split(";")
-        case property
+
+        parser, attr, occurrences = *send(parser, property, value)
+
+        case attr
```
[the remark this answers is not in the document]
iainbeeston: Sorry for the long wait, I've refactored this now to get rid of the extra case statement. Please let me know what you think.

PR #480, opened 2020-03-06
CL2-5023 Drop support for old ruby and ruby rails versions (closed PR #459)

sebastienhoorens:
The fork we used was deleted, breaking our backend. Here is our own reconstructed version of that fork.


sebastienhoorens:
Please ignore this 😅 (made the wrong PR)

PR #484, opened 2020-09-06
Fix warning messages

vkhalzov:
Fix warning messages.
Not as elegant as it was before, but no warnings.


seejohnrun:
Thank you! I appreciate your time here / I've used `delegate` to silence the 2.8 warnings which I think reads a bit cleaner

PR #487, opened 2020-10-28
[Bugfix] Weekly interval rules are sometimes calculated incorrectly

jakebrady5:
The spec covers the context through which this bug was uncovered. Essentially generating `occurrences_between` for a relatively small date range was yielding false occurrences for weekly interval rules.

I observed that `@time = @start_time if @time < @start_time` was set to happen only when `@start_time` was falsey. Iterations past the first for a given rule in `Schedule#next_time` [here](https://github.com/seejohnrun/ice_cube/blob/master/lib/ice_cube/schedule.rb#L451) in situations similar to the spec would pass in a `time` that was less than `start_time`.

Since `@start_time` was already defined in the rule, this `@time` value would be passed long for [validation](https://github.com/seejohnrun/ice_cube/blob/master/lib/ice_cube/validated_rule.rb#L147).

Most rules will then [return](https://github.com/seejohnrun/ice_cube/blob/master/lib/ice_cube/validations/weekly_interval.rb#L31) immediately in this case, not applying their validation.

Any feedback is welcome!

### DIFF
--- a/spec/examples/weekly_rule_spec.rb
+++ b/spec/examples/weekly_rule_spec.rb
@@ -292,6 +292,19 @@ module IceCube
         expect(schedule.next_occurrence(sample[2] - 1)).to eq(sample[2])
       end
 
+      it 'should respect weekly intervals within narrow occurrence ranges' do
+        start_time = Time.utc(2020, 10, 27, 7, 0, 0)
+        schedule = Schedule.new(start_time, end_time: start_time + ONE_HOUR)
+        occurrence_start = Time.utc(2020, 11, 5, 0, 0, 0)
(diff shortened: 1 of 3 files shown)


spec/examples/weekly_rule_spec.rb
```diff
@@ -288,6 +288,19 @@ module IceCube
         expect(schedule.next_occurrence(sample[2] - 1)).to eq(sample[2])
       end
 
+      it 'should respect weekly intervals within narrow occurrence ranges' do
```
pacso, line 291:
```suggestion
      it "should respect weekly intervals within narrow occurrence ranges" do
```

PR #489, opened 2020-11-17
Fix I18n.load_path ordering conflicts with rails_i18n

brianswko:
Addresses https://github.com/seejohnrun/ice_cube/pull/432 and https://github.com/seejohnrun/ice_cube/issues/431

Since `ice_cube`'s translation files are different from the defaults in `rails-i18n`, the translations are incorrect when `ice_cube` gets loaded.

For example, in french for the abbreviated form of January:
`rails-18n` uses `jan.` https://github.com/svenfuchs/rails-i18n/blob/0cb422dc111b2dfa61a19b4740d04e72fb082d17/rails/locale/fr.yml#L23
`ice_cube` uses `Jan` https://github.com/seejohnrun/ice_cube/blob/d4443a29de12d277e04fede34df4d0e626a9634e/config/locales/fr.yml#L124

When `ice_cube` gets loaded, it overrides the default resulting in incorrect formatting. This change checks for the presence of `rails_i18n` in the load path. If it exists, it inserts `ice_cube`'s files before them so that conflicts get overridden by the defaults. If `rails_i18n` is not in use, then they get added to the end of the load path.


brianswko:
I've looked through the linked discussions and haven't really seen any movement on them recently. We've solved this for our own application by monkey patching `IceCube::I18n` in this exact way, let me know what you think!

timfsw:
@brianswko Thanks for the fix - we issued the same problem and your solution worked great.

Are there any plans to merge it into master?

joshRpowell:
poke @seejohnrun ☝🏻

Faq:
So looks like this PR stuck only because missing record for CHANGELOG?

krtschmr:
Yes. welcome to open source development 😢

Issue #514, opened 2021-10-27
Inconsistent daily schedule with interval behaviour between version 0.16.3 and 0.16.4


unrooty-infinum:
Hello! First of all thanks for the gem, you did a great job!

I'm facing an issue with daily schedule with interval after `ice_cube` update from version `0.16.3` to version `0.16.4`.

According the following reproducion scripts you can see that version `0.16.3` output differs from version `0.16.4` output.

Is it expected behaviour? And which version output is correct? 

## Environment
Ruby: 3.0.2p107 (2021-07-07 revision 0db68f0233) [arm64-darwin20]
OS: MacOS Big Sur 11.6

## Version 0.16.3 behaviour reproducion script
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'ice_cube', '0.16.3'
end

require 'ice_cube'

interval = 4

rule = IceCube::Rule.daily(interval).hour_of_day(9).minute_of_hour(0).second_of_minute(0)

start_time = Time.new(2021, 3, 28, 10, 0, 0, 'UTC') # 2021-03-28 10:00:00 UTC
end_time = Time.new(2021, 4, 7, 23, 59, 59, 'UTC') # 2021-04-07 23:59:59 UTC

schedule = IceCube::Schedule.new(start_time) { |s| s.add_recurrence_rule(rule) }

schedule.occurrences(end_time)

# Output
# 2021-03-29 09:00:00 UTC
# 2021-04-02 09:00:00 UTC
# 2021-04-06 09:00:00 UTC
```
## Version 0.16.4 behaviour reproducion script
```ruby
require 'bundler/inline'

gemfile do
  source 'https://rubygems.org'
  gem 'ice_cube', '0.16.4'
end

require 'ice_cube'

interval = 4

rule = IceCube::Rule.daily(interval).hour_of_day(9).minute_of_hour(0).second_of_minute(0)

start_time = Time.new(2021, 3, 28, 10, 0, 0, 'UTC') # 2021-03-28 10:00:00 UTC
end_time = Time.new(2021, 4, 7, 23, 59, 59, 'UTC') # 2021-04-07 23:59:59 UTC

schedule = IceCube::Schedule.new(start_time) { |s| s.add_recurrence_rule(rule) }

schedule.occurrences(end_time)

# Output
# 2021-04-01 09:00:00 UTC
# 2021-04-05 09:00:00 UTC

PR #523, opened 2022-03-02
Fix daily schedule bug

pacso:
This PR reverts the changes introduced in #452 which caused the issue detailed in #514


pacso:
@seejohnrun - using `.occurrences` as per the new test added [here](https://github.com/ice-cube-ruby/ice_cube/pull/523/files#diff-a2d3993d84af8f47303f4214174ddfc13cfe4fb4af129af32d04bf55a97fdf45R365), you can see that the schedule is adding the correct first occurrence.

Do you have any insight into why the existing test below it might be failing?

PR #528, opened 2022-06-08
Fix `DEPRECATION WARNING: TimeWithZone#to_s(:short)`

iainbeeston:
Rails has deprecated passing a parameter to `to_s` and
now expects `to_formatted_s` to be used instead.

This is the full error I see:

    DEPRECATION WARNING: TimeWithZone#to_s(:short) is deprecated. Please use TimeWithZone#to_fs(:short) instead. (called from to_s at /ice-cube-ruby/ice_cube/lib/ice_cube/occurrence.rb:89)

### DIFF
--- a/lib/ice_cube/occurrence.rb
+++ b/lib/ice_cube/occurrence.rb
@@ -86,7 +86,7 @@ module IceCube
     #
     def to_s(format = nil)
       if format && to_time.public_method(:to_s).arity != 0
-        t0, t1 = start_time.to_s(format), end_time.to_s(format)
+        t0, t1 = start_time.to_formatted_s(format), end_time.to_formatted_s(format)
       else
         t0, t1 = start_time.to_s, end_time.to_s
       end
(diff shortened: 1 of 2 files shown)


adfoster-r7:
Looks like this is now an error in newer versions of Active Support:

```
Failures:

  1) IceCube::Occurrence to_s accepts a format option to comply with ActiveSupport
     Failure/Error: expect(occurrence.to_s(:short)).to eq time_now.to_s(:short)
     
     ArgumentError:
       wrong number of arguments (given 1, expected 0)
     # /Users/user/.rvm/gems/ruby-3.0.2@pro/gems/activesupport-7.1.3/lib/active_support/time_with_zone.rb:200:in `to_s'
     # ./spec/examples/occurrence_spec.rb:31:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:76:in `block (3 levels) in <top (required)>'
     # ./spec/spec_helper.rb:75:in `block (2 levels) in <top (required)>'
```

After applying these changes locally, they pass though :+1:

pacso:
Fixed in #449

PR #536, opened 2022-11-16
Fix some failing round trips

boutil:
In some Debian environment two of the round trips in to_yaml_spec.rb were failing (those using `.month_of_year` and `day_of_year`. Applying the same fix as the one used in [4bf17183be02f8efe17b98c401af40db48a931dc](https://github.com/ice-cube-ruby/ice_cube/commit/4bf17183be02f8efe17b98c401af40db48a931dc) to fix #483 solved the issue.

This patch proposes that corresponding change to `/spec/examples/to_yaml_spec.rb`.

PR #537, opened 2022-12-12
Fix the sensitivity to DST on a few tests

nehresma:
There are 2 tests that are failing currently due to being sensitive to daylight savings time changes.


nehresma:
Sorry, this is a dupe of #536.  Closing.

Issue #547, opened 2023-12-02
Update RubyGems/gemspec Links


chilliphili:
The Homepage, Wiki and Review Changes links on https://rubygems.org/gems/ice_cube are broken as they point to the original repository for this gem. This issue is to update the references in the gemspec file to point to this repository as specified by https://guides.rubygems.org/specification-reference.

pacso:
Was fixed in #553

PR #548, opened 2023-12-02
Update gemspec links

chilliphili:
This commit updates the gemspec links to point to ice-cube-ruby/ice_cube which will fix the broken links on https://rubygems.org/gems/ice_cube.

https://github.com/ice-cube-ruby/ice_cube/issues/547


pacso:
No longer required as was updated in #553

PR #551, opened 2024-04-24
weekly rule DST issue

larskuhnt:
Hi,

I added some specs for the following edge case where the calculation of the next occurrences returns faulty values.
The base of this issue is `IceCube::WeeklyRule.realign` when an occurrence happens within a DST time switch. For example on `26th April 2024 00:00` -> `01:00`. Realign will change the start_time of e.g. `22th May 2022 00:20` to `26th April 2024 01:20` and then the starting hour is wrong. The result is occurrences with the time `01:20` instead of `00:20`. 

It would be great if someone could guide me to the place where this can get fixed. 

Thanks
Lars

### DIFF
--- a/spec/examples/schedule_spec.rb
+++ b/spec/examples/schedule_spec.rb
@@ -390,6 +390,35 @@ describe IceCube::Schedule do
       expect(next_hours).to eq [Time.utc(2014, 1, 2, 0o0, 34, 56),
         Time.utc(2014, 1, 2, 0o1, 34, 56)]
     end
+
+    context "Cairo timezone" do
+      let(:schedule) do
+        IceCube::Schedule.from_yaml("---\n:start_time:\n  :time: 2022-05-05 22:20:00.000000000 Z\n  :zone: Africa/Cairo\n:end_time:\n  :time: 2022-05-06 21:40:00.000000000 Z\n  :zone: Africa/Cairo\n:rrules:\n- :validations:\n    :day:\n    - 5\n  :rule_type: IceCube::WeeklyRule\n  :interval: 1\n  :week_start: 1\n:rtimes: []\n:extimes: []\n")
+      end
+      
+      it "has the correct start time" do
+        expect(schedule.start_time.iso8601).to eq("2022-05-06T00:20:00+02:00")
+      end
+
+      it "calculates the corret occurrences from 2024-04-24" do
+        ref_time = Time.utc(2024, 4, 24, 12, 0, 0)
+        occurrences = schedule.next_occurrences(3, ref_time)
+        expect(occurrences.map(&:iso8601)).to eq([
+          "2024-04-26T01:20:00+03:00",
+          "2024-05-03T00:20:00+03:00",
+          "2024-05-10T00:20:00+03:00"
+        ])
+      end
+
+      it "calculates the corret occurrences from 2024-04-21" do
+        occurrences = schedule.next_occurrences(3, Time.utc(2024, 4, 21, 12, 0, 0))
+        expect(occurrences.map(&:iso8601)).to eq([
+          "2024-04-26T01:20:00+03:00",
+          "2024-05-03T00:20:00+03:00",
+          "2024-05-10T00:20:00+03:00"
+        ])
+      end
+    end
   end
 
   describe :next_occurrence do
--- a/spec/examples/weekly_rule_spec.rb
+++ b/spec/examples/weekly_rule_spec.rb
@@ -390,5 +390,40 @@ module IceCube
         end
       end
     end
+
+    context "when start time is within a timezone shift in Africa/Cairo timezone", system_time_zone: "UTC" do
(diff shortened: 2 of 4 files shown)


spec/examples/weekly_rule_spec.rb
```diff
@@ -390,5 +390,40 @@ module IceCube
...
+
+      it "calculates the correct time from 2024-04-26 00:20:01 Africa/Cairo" do
+        expect(rule.next_time(cairo_tz.parse("2024-04-26 00:20:01"), start_time, nil).iso8601).to eq("2024-05-03T00:20:00+03:00")
+      end
+    end
+
+    describe :realign do
+      let(:cairo_tz) { ActiveSupport::TimeZone["Africa/Cairo"] }
+      let(:utc_tz) { ActiveSupport::TimeZone["UTC"] }
+      let(:start_time) { cairo_tz.parse("2022-05-22 00:20:00") }
```
pacso, line 416: I assume you've picked this date and timezone because they've been faffing with using and not using summer time?

Would it be more obvious to use a less contrived example? It took me a while to realise why the offset was changing from +2 to +3?

spec/examples/weekly_rule_spec.rb
```diff
@@ -390,5 +390,40 @@ module IceCube
...
+      let(:time) { utc_tz.parse("2024-04-24 12:00:00") }
+      let(:rule) { Rule.weekly(1, :monday).day(:friday) }
+
+      subject { rule.realign(time, start_time) }
+
+      it { puts cairo_tz.parse("2024-04-26T00:20:00") }
```
pacso, line 422: We probably don't want this

spec/examples/schedule_spec.rb
```diff
@@ -390,6 +390,35 @@
...
+        ref_time = Time.utc(2024, 4, 24, 12, 0, 0)
+        occurrences = schedule.next_occurrences(3, ref_time)
+        expect(occurrences.map(&:iso8601)).to eq([
+          "2024-04-26T00:20:00+03:00",
+          "2024-05-03T00:20:00+03:00",
+          "2024-05-10T00:20:00+03:00"
+        ])
+      end
+
+      it "calculates the corret occurrences from 2024-04-21" do
```
pacso, line 413: Can you change one of these such that the set of occurrences span the transition to DST?

pacso:
Hi @larskuhnt,

Thank you for your contribution ... this looks great! Just a couple of comments, but otherwise looks spot on. 👍🏻

pacso:
Don't worry about fixing the linting issues ... that is resolved on #552 which I'll merge first.

I'm just waiting on @seejohnrun updating my access so I can actually fix things.

larskuhnt:
Hi @pacso,
your comment https://github.com/ice-cube-ruby/ice_cube/pull/551#discussion_r1580097086 was very helpful as it showed a flaw in my approach to fix this issue. I added some more specs now and decided to realign the time to a day before the DST switch for this special case to keep the timezone information. The only issue with this approach could be when the DST switch happens on a monday/sunday so that the realigned time is in the previous week. But I am not sure about that.

pacso:
@larskuhnt - could you please pull `master` and see whether this fixes all of the tests?

larskuhnt:
@pacso I merged `master` and the specs are green on my side.

pacso:
@larskuhnt - Could you resolve the linting errors?

larskuhnt:
@pacso all checks pass now

pacso:
Amazing, thank you! I'll do a proper review either at lunch or after work and get this merged in 👍🏻

lib/ice_cube/rules/weekly_rule.rb
```diff
@@ -35,7 +35,17 @@ def realign(step_time, start_time)
...
+      realigned_time = time.to_time
+      # when the realigned time is in a different hour, it means that
+      # time falls to the DST switch timespan. In this case, we need to
+      # move the time back by one day to ensure that the hour stays the same
+      # WARNING: this could not work if the DST change is on a monday
+      # as the realigned time would be moved to the previous week.
+      if realigned_time.hour != start_time.hour
+        time.add(:day, -1)
+        realigned_time = time.to_time
+      end
```
pacso, line 47: Can we not just detect the difference between the `start_time.hour` and the `realigned_time.hour` and make a 60 minute adjustment to the realigned_time to correct for any DST changes? That would avoid any issues with switching weeks, etc.

larskuhnt: This case can only happen when the realigned time changed the hour in the internals of `Time` because of a DST change, so -1 day will always be a time which does not fall into the DST change time imho.

I also see that this is not the ideal solution, it would be better to somehow "fix" the hour to the start time hour of the schedule somewhere in `ValidatedRule#next_time` but I did not find a way to access the start time of the schedule in `ValidatedRule`. Maybe you have an idea to that approach.

Regarding the edge case: that was just a guess based on the documentation for the message. I don't really understand the described edge case and thought that the movement to the previous day could somehow trigger that case again.

PR #554, opened 2024-05-10
Validation input checking

### LINKED ISSUE #469: Infinite loop with day_of_week
Essentially doing this makes an infinite loop: `day_of_week(monday: [0])`

Full reproduction test case:

```ruby
schedule = IceCube::Schedule.new(Time.zone.now)
schedule.add_recurrence_rule(IceCube::Rule.daily.day_of_week(monday: [0]))
schedule.first
```

pacso:
Fixes #469


lib/ice_cube/validations/day_of_week.rb
```diff
@@ -15,6 +15,8 @@ class Validation
       attr_reader :day, :occ
 
       def initialize(day, occ)
+        raise ArgumentError, "Invalid day_of_week occurrence: #{occ}" if occ == 0 || occ > 5 || occ < -5
```
pacso, line 18: @seejohnrun - I'm curious why input validations like this weren't added to all of the validation classes. Was it a deliberate decision not to do this?

PR #561, opened 2024-09-23
Fix Syntax Error in ice_cube gem for Ruby 3.0+ Compatibility

codxse:
This pull request addresses a syntax error in the ice_cube gem that occurs when using Ruby 3.0 or later. The error is caused by the use of a shorthand syntax for block parameters that is not supported in earlier Ruby versions.

Current Error

```
SyntaxError:
       /usr/local/bundle/bundler/gems/ice_cube-db4715e37368/lib/ice_cube/schedule.rb:163: syntax error, unexpected ')', expecting local variable or method
           def each_occurrence(&)
                                ^
       /usr/local/bundle/bundler/gems/ice_cube-db4715e37368/lib/ice_cube/schedule.rb:164: syntax error, unexpected ')'
       ...rate_occurrences(start_time, &).to_a
       ...                              ^
       /usr/local/bundle/bundler/gems/ice_cube-db4715e37368/lib/ice_cube/schedule.rb:502: syntax error, unexpected `end', expecting end-of-input
```

This PR modifies the each_occurrence method to use the traditional block syntax, which is compatible with both older and newer Ruby versions. The changes are as follows:

```
# Old code
def each_occurrence(&)
  generate_occurrences(start_time, &).to_a
end

# New code
def each_occurrence(&block)
  generate_occurrences(start_time, &block).to_a
end
```

This change maintains the functionality while ensuring compatibility across Ruby versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -33,7 +33,7 @@ jobs:
     - uses: actions/checkout@v4
     - uses: ruby/setup-ruby@v1
       with:
-        ruby-version: '3.1'
+        ruby-version: '3.4'
         bundler-cache: true
     - run: bundle exec standardrb
   test:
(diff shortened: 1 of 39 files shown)


pacso:
This was fixed in #562

PR #562, opened 2024-11-06
Add backwards compatibility to Ruby 2.6

scpike:
This removes use of the "anonymous block argument" feature which is only supported in Ruby 3.1+ so that the library stays backwards compatible to earlier ruby versions.

### DIFF
--- a/.github/workflows/tests.yaml
+++ b/.github/workflows/tests.yaml
@@ -43,7 +43,7 @@ jobs:
     strategy:
       matrix:
         rails: ['6.1', '7.0', '7.1']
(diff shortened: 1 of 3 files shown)


pacso:
@seejohnrun - what are your thoughts on this? Do we want to actively support version of Ruby / Rails which are end-of-life? If we do, where do we draw the line?

pacso:
Yep - it's a fair comment.

I've updated the PR to pass against all of the versions back to Rails 6.1 and Ruby 2.6. I'll review it again in the morning when I'm not so tired and get it merged.

PR #566, opened 2025-01-13
Support Ruby 3.0

T-800:
Hi,

I tried to use Ruby 3.0 with this gem and I found that some error syntax error which are fine on Ruby 2.7. Please help to review and merge it if you find it does make sences.

Thanks.


T-800:
Duplicated with https://github.com/ice-cube-ruby/ice_cube/pull/562 and https://github.com/ice-cube-ruby/ice_cube/pull/561

PR #569, opened 2025-06-17
Better preserve timezones in ical

sergioisidoro:


pacso:
Hello,

Thank you for your work on this! Am keen to keep this project alive. I will try to put some time aside this weekend to review this PR properly and see if there's a way we can get this merged.

wwahammy:
> Hello,
> 
> Thank you for your work on this! Am keen to keep this project alive. I will try to put some time aside this weekend to review this PR properly and see if there's a way we can get this merged.

Thank you for all of the work you do @pacso! Totally get that you're doing this on the side and that's a lot of work. If you're open to it, it might make sense to work together on developing a path forward that spreads the workload more so you're not responsible for so much. We're all benefiting so we should all pitch-in after all.

lib/ice_cube/builders/ical_builder.rb
```diff
@@ -35,10 +35,22 @@ def self.ical_utc_format(time)
 
     def self.ical_format(time, force_utc)
       time = time.dup.utc if force_utc
+
+      # Keep timezone. strftime will serializer short versions of time zone (eg. EEST),
+      # which are not reversivible, as there are many repeated abbreviated zones. This will result in
```
sergioisidoro, line 40:
```suggestion
      # which are not reversible, as there are many repeated abbreviated zones. This will result in
```

lib/ice_cube/builders/ical_builder.rb
```diff
@@ -35,10 +35,22 @@ def self.ical_utc_format(time)
...
+        tz_id = time.time_zone.name
+        return ";TZID=#{tz_id}:#{IceCube::I18n.l(time, format: "%Y%m%dT%H%M%S")}" # local time specified"
+      end
+
       if time.utc?
         ":#{IceCube::I18n.l(time, format: "%Y%m%dT%H%M%SZ")}" # utc time
       else
-        ";TZID=#{IceCube::I18n.l(time, format: "%Z:%Y%m%dT%H%M%S")}" # local time specified
+        # Convert to UTC as TZID=+xxxx format is not recognized by JS libraries
+        warn "IceCube: Time object does not have timezone info. Assuming UTC: #{caller(1..1).first}"
```
sergioisidoro, line 51:
```suggestion
        # Convert to UTC as TZID=+xxxx format is not recognized by JS libraries
        warn "IceCube: Time object does not have timezone info. Coercing into UTC: #{caller(1..1).first}"
```

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -26,6 +33,22 @@ def self.schedule_from_ical(ical_string, options = {})
       Schedule.from_hash data
     end
 
+    def self.deserialize_time_with_tzid(time_value, tzid)
```
sergioisidoro, line 36: This should maybe be moved into the TimeUtil class since it's a wrapper for the `TimeUtil.deserialize_time`

lib/ice_cube/parsers/ical_parser.rb
```diff
@@ -26,6 +33,22 @@ def self.schedule_from_ical(ical_string, options = {})
...
+        # No TZID, use standard deserialization
+        TimeUtil.deserialize_time(time_value)
+      else
+        # TZID is a timezone name - Assume it's a valid timezone in a try-catch block
+        begin
+          TimeUtil.deserialize_time({time: time_value, zone: tzid})
+        rescue ArgumentError
+          # If the timezone is invalid, fall back to standard deserialization
+          # Perhaps we want to log this?
+          TimeUtil.deserialize_time(time_value)
```
sergioisidoro, line 47: Should we issue a warning here as well?

sergioisidoro (commented):
I self reviewed the PR :) there were a few typos and a small nitpick. This work was done a bit in a hurry to fix our bugs with the lost time in serialization.

Let me know if there's anything else that needs attention on this PR.

sergioisidoro:
Status? 

@wwahammy are you still open to help maintaining a fork?

wwahammy:
> Status?
> 
> @wwahammy are you still open to help maintaining a fork?

Hi @sergioisidoro, yes I'm still open to it!

lib/ice_cube/builders/ical_builder.rb
```diff
@@ -36,24 +36,45 @@ def self.ical_utc_format(time)
...
+    end
+
+    # A schedule built from plain Time objects hits this on every serialization,
+    # so warn once per process rather than once per occurrence.
+    def self.warn_missing_time_zone
+      return if @missing_time_zone_warned
+
+      @missing_time_zone_warned = true
+      warn "IceCube: Time object does not have timezone info. Coercing into UTC: #{caller(2..2).first}"
+    end
```
wwahammy, line 76: ~~I think this warn change should be its own PR, it's not really related to preserving timezones in ical.~~ EDIT: Oops, nevermind, I realized you added the warn in this PR originally.  Again, 7:30 in the morning :smile: 


---


Additionally, it's 7:30 in the morning so I might be reading this wrong but are we sure this is actually going to warn once per process? Is there a situation where the class gets reloaded or there's some sort of race condition on this?

sergioisidoro: Yes, I added this warning because it was one of the reasons I made this PR - it defaulted to falling back to UTC without anything surfaced to the dev. If I would have seen this in the initial test runs, it would have been clear that the behaviour was not expected. I think this also serves a bit as a mitigation for the breaking change.

Good question. This change was introduced by claude code in the lastest round of reviewing (which fixed the ActiveRecord dependency issue). I previosuly was issuing warning per instance, which could easily overwhelm the logs. First question is -- is it enough to warn once, to allow the dev to know of this behaviour? Or is issuing a warning per instance actually desired? 

About the per process, I had to search this. Ruby classes are instances of a Class, so that instance variable `missing_time_zone_warned` will be set once per class when it is loaded. I assume that a reload will re-instantiate that `Class` instance 🤔 It solves the problem of spamming the logs, but the comment may be factually incorrect.

